PR #29, opened 2019-01-10
Make update seq id optional

atrakh:
Alters the `updateSequenceId` parameter to be optional. If not provided (or set to a number < 0), data will always be updated. If provided, data will only be updated if the existing `updateSequenceId` is less than the new.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change log
 
 All notable changes to the LaunchDarkly git-flag-parser will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org).
+## [0.1.1] - 2019-01-10
+### Changed
+- `updateSequenceId` is now an optional parameter. If not provided (or set to a number < 0), data will always be updated. If provided, data will only be updated if the existing `updateSequenceId` is less than the new `updateSequenceId`.
 
 ## [0.1.0] - 2019-01-02
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Change log
 
 All notable changes to the LaunchDarkly git-flag-parser will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org).
+
 ## [0.1.1] - 2019-01-10
 ### Changed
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to the LaunchDarkly git-flag-parser will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org).
 
+## [0.1.1] - 2019-01-10
+### Changed
+- `updateSequenceId` is now an optional parameter. If not provided (or set to a number < 0), data will always be updated. If provided, data will only be updated if the existing `updateSequenceId` is less than the new `updateSequenceId`.
```
vlymar: whats the rationale for making negative numbers always trigger updates?

README.md
```diff
@@ -31,7 +31,7 @@ The following options are available to the program:
 | `dir` | Path to existing checkout of the git repo. If a cloneEndpoint is provided, this option is not required. |  | only if `cloneEndpoint` is not set |
 | `exclude` | A regular expression defining the files and directories which the flag parser should exclude. |  | no |
 | `projKey` | A LaunchDarkly project key. |  | yes |
-| `updateSequenceId` | An integer representing the order number of code reference updates. Used to version updates across concurrent executions of the flag parser. Examples: the time a `git push` was initiated, CI build number, the current time. |  | yes |
+| `updateSequenceId` | An integer representing the order number of code reference updates. Used to version updates across concurrent executions of the flag parser. If not provided (or set to a number < 0), data will always be updated. If provided, data will only be updated if the existing `updateSequenceId` is less than the new `updateSequenceId`. Examples: the time a `git push` was initiated, CI build number, the current time. |  | no |
```
vlymar: i vote we update "the current time" to "current unix timestamp" to make it clear that someone can't provide HH:MM timestamps

CHANGELOG.md
atrakh: This is sort of a workaround to account for command-line options requiring a default value of the underlying type. Since we can't give pointer values to the `flag` package, using negative numbers to indicate that there is no update sequence id. Another alternative here would be to just have the underlying type of the option be a string, but that would affect the auto-generated "help" documentation of the command line tool

vlymar: you can't specify that the command-line option is optional? if its optional wouldn't it require a default value?

PR #30, opened 2019-01-11
Don't grep for flags with short keys.

vlymar:
Flags with really short keys like `a` or `ab` end up generating many
false positives, drowning out real code references with noise. For now
we will just not support those flags for code references.

I wonder if we should allow users to override this minimum with a CLI param? I'm leaning towards waiting for if/when someone requests that feature.

TODO:
- [ ] documentation


atrakh:
Approved pending green checks on ci

PR #33, opened 2019-01-14
Disable cloneEndpoint argument

atrakh:
We've noticed a bit of confusion between the dir and cloneEndpoint arguments for the parser, and haven't really gained any value out of cloneEndpoint, other than manual testing early on. Ideally, users should have their repo cloned into their workspace before running the parser, and it shouldn't be our concern to do it for them.

So, I recommend we remove cloneEndpoint for now, and add it back in later (code will be commented out, so there will be little cost), if we decide we need it. 

I'll look into our existing wrappers to see if we can get away with removing `repoHead` as well.

tl;dr: If `cloneEndpoint` is not useful, we shouldn't release it in v1, because we would need to support it.


README.md
```diff
@@ -32,7 +32,7 @@ A number of command-line arguments are available to the parser, some optional, a
 | Option | Description |
 |-|-|
 | `accessToken` | LaunchDarkly [personal access token](https://docs.launchdarkly.com/docs/api-access-tokens) with writer-level access, or access to the `code-references` [custom role](https://docs.launchdarkly.com/v2.0/docs/custom-roles) resource |
-| `dir` | Path to existing checkout of the git repo. If a cloneEndpoint is provided, this option is not required. |
+| `dir` | Path to existing checkout of the git repo. |
```
vlymar: also mentioned in the documentation readme, but this should specify "Absolute path"

PR #34, opened 2019-01-15
Add a line length limit.

vlymar:
After 500 characters, truncate lines.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to the LaunchDarkly git-flag-parser will be documented in th
 ### Changed
 - `updateSequenceId` is now an optional parameter. If not provided, data will always be updated. If provided, data will only be updated if the existing `updateSequenceId` is less than the new `updateSequenceId`.
 - Flags with keys shorter than 3 characters are no longer supported.
+- lines are truncated after 500 characters.
 
 ## [0.1.0] - 2019-01-02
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,10 +2,10 @@
 
 All notable changes to the LaunchDarkly git-flag-parser will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org).
 
-## [0.1.1] - 2019-01-11
```
vlymar: intentionally removed. the convention I like is to just keep building up the changelog for whats in master, but not add a version or date until we're ready to release.

parse/parse.go
```diff
@@ -401,3 +402,16 @@ func fatal(msg string, err error) {
...
 }
+
+// Truncate lines to prevent sending over massive hunks, e.g. a minified file.
+func truncateLine(line string) string {
+	// len(line) returns number of bytes, not num. characters, but it's a close enough
+	// approximation for our purposes
+	if len(line) > maxLineCharCount {
+		// convert to rune slice so that we don't truncate multibyte unicode characters
+		runes := []rune(line)
+		return string(runes[0:maxLineCharCount])
```
atrakh: Should we put some character at the end here to indicate that the line has been terminated?

vlymar: done in 972b264

PR #35, opened 2019-01-16
Limit number of files, and number of hunked lines per file/flag pair.

vlymar:
I've tested all these limits by cranking them down and running the command against SupportService.

Note that we can't actually calculate the upper bound payload limit from these numbers because size will depend on number of flags.

https://app.clubhouse.io/launchdarkly/story/27365/add-a-minimum-character-limit-for-flags-scanned-by-the-git-flag-parser-and-other-limits

### DIFF
--- a/parse/parse.go
+++ b/parse/parse.go
@@ -394,7 +394,6 @@ func buildHunksForFlag(projKey, flag string, flagReferences []*list.Element, fil
 			currentHunk.Lines = hunkStringBuilder.String()
 			hunks = append(hunks, currentHunk)
 			previousHunk = &hunks[len(hunks)-1]
-
 		}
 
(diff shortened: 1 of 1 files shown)


parse/parse.go
```diff
@@ -232,8 +241,31 @@ func (g grepResultLines) makeReferenceHunksReps(projKey string, ctxLines int) []
 
 	aggregatedGrepResults := g.aggregateByPath()
 
+	if len(aggregatedGrepResults) > maxFileCount {
+		log.Info("number of files containing code references exceeded limit",
```
atrakh: Should these be warnings?

vlymar: good point

vlymar: will be done in logging PR

PR #46, opened 2019-01-18
Infer current branch checkout, remove repoHead arguments

atrakh:
This PR removes the `repoHead` command line argument in effort to simplify the utility. 

Added new logic to infer the branch name using a special `git rev-parse` flag.

### DIFF
--- a/build/package/circleci/orb.yml
+++ b/build/package/circleci/orb.yml
@@ -51,7 +51,6 @@ commands:
               -repoType=${LD_REPO_TYPE} \
               -repoUrl=${LD_REPO_URL} \
               -repoName=${CIRCLE_PROJECT_REPONAME} \
-              -repoHead=${CIRCLE_BRANCH} \
               -updateSequenceId=${CIRCLE_BUILD_NUM} \
               -defaultBranch=${LD_DEFAULT_BRANCH} \
               -commitUrlTemplate=${LD_COMMIT_URL_TEMPLATE} \
--- a/build/package/circleci/orb.yml
+++ b/build/package/circleci/orb.yml
@@ -73,7 +73,7 @@ jobs:
         type: integer
         default: -1
       exclude:
-        description:  "A regular expression (PCRE) defining the files, file types, and directories which the job should exclude. Partial matches are allowed. Examples: `vendor/`, `vendor/.*`, `.*\\.css`"
+        description:  "A regular expression (PCRE) defining the files, file types, and directories which the job should exclude. Partial matches are allowed. Examples: `vendor/`, `\.css`, `vendor/|\.css`"
         type: string
         default: ""
       repo_type:
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -9,8 +9,7 @@ import (
 
 type Git struct {
 	Workspace string
-	Head      string
-	RepoName  string
+	// Head      string
 }
 
 /*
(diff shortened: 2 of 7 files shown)


pkg/parse/parse.go
```diff
@@ -64,29 +64,20 @@ func Parse() {
...
-	// 	if err != nil {
-	// 		fatal("Failed to create temp directory for repo clone", err)
-	// 	}
-	// 	defer os.RemoveAll(dir)
-	// 	cmd.Workspace = dir
-	// 	err = cmd.Clone(endpoint)
-	// 	if err != nil {
-	// 		fatal("Unable to clone repo", err)
-	// 	}
-	// }
```
atrakh: Decided it's pretty safe to remove this commented out code path. If we ever want this functionality back, we can do a git revert on the commit that commented this code out.

pkg/parse/parse.go
```diff
@@ -64,29 +64,20 @@ func Parse() {
...
+	cmd := git.Git{Workspace: o.Dir.Value()}
 
-	headSha, err := cmd.RevParse()
+	currBranch, err := cmd.BranchName()
+	if err != nil {
+		log.Error.Fatalf("error parsing git branch name: %s", err)
+	} else if strings.ToUpper(currBranch) == "HEAD" || currBranch == "" {
+		log.Error.Fatalf("error parsing git branch name: git repo at %s must be checked out to a valid branch", cmd.Workspace)
```
atrakh: This errors out if you actually named your branch `head`, but I don't think that's a huge issue

internal/git/git.go
```diff
@@ -9,8 +9,7 @@ import (
 
 type Git struct {
 	Workspace string
-	Head      string
-	RepoName  string
+	// Head      string
```
vlymar: why is this commented out?

pkg/parse/parse.go
```diff
@@ -177,12 +168,7 @@ func getFlags(ldApi ld.ApiClient) ([]string, error) {
 }
 
 func (b *branch) findReferences(cmd git.Git, flags []string, ctxLines int, exclude *regexp.Regexp) (grepResultLines, error) {
-	err := cmd.Checkout()
```
vlymar: sanity check - do github/bitbucket give us the repo with the right branch already checked out?

pkg/parse/parse.go
vlymar: why is a branch name of `head` a problem?

build/package/circleci/orb.yml
```diff
@@ -51,7 +51,6 @@ commands:
               -repoType=${LD_REPO_TYPE} \
               -repoUrl=${LD_REPO_URL} \
               -repoName=${CIRCLE_PROJECT_REPONAME} \
-              -repoHead=${CIRCLE_BRANCH} \
```
vlymar: we can assume that in CLI situations people will manually have the desired branch checked out, and hopefully github/bitbucket pre-checkout the correct branch, but does circle provide the repo with the correct branch checked out?

atrakh: Yes, circle checks out to the correct branch.

pkg/parse/parse.go
atrakh: Yes, I checked bb and github, and they appear to give us the correct branch checkout. I'll make a release candidate out of this branch and double check as well.

pkg/parse/parse.go
atrakh: It's the output returned by `git rev-parse --abbrev-ref` if you're in a detached state. I suppose we could detect this as a problem earlier on so this check doesn't happen in `parse.go`

vlymar: yeah is there some way to just make sure the repo isn't in a detached state? that being said, i guess this is a nice to have. I could just see a user being surprised that we'd have this restriction. We can also always add support for branches named `head` later w/o a backwards incompatible change

atrakh:
filed: https://app.clubhouse.io/launchdarkly/story/30379/ld-find-code-refs-add-extra-checks-for-detached-state-to-support-branches-named-head

PR #49, opened 2019-01-23
Prepare 0.3.0

atrakh:


atrakh:
> we should probably hold this until we have the new contextLines default and updated documentation merged in right?

resolved

PR #51, opened 2019-01-23
Add debug command-line argument

atrakh:
Adds the optional `debug` command line argument. If enabled, enables debug logging. The pretty-printed table of flag keys to ref counts now requires debug logging to be printed.

### DIFF
--- a/internal/options/options.go
+++ b/internal/options/options.go
@@ -211,7 +211,7 @@ func GetLDOptionsFromEnv() (map[string]string, error) {
 }
 
 func GetDebugOptionFromEnv() (bool, error) {
-	debug := os.Getenv("debug")
+	debug := os.Getenv("LD_DEBUG")
 	if debug == "" {
 		return false, nil
 	}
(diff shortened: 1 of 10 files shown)


pkg/parse/parse.go
```diff
@@ -60,13 +60,6 @@ type branch struct {
 }
 
 func Parse() {
-	err, cb := o.Init()
-	if err != nil {
-		log.Error.Printf("could not validate command line options: %s", err)
-		cb()
-		os.Exit(1)
-	}
-
```
atrakh: Moved logging initialization up to the actual command line tools / build packages since log configuration is variable now.

If logging isn't configured when `Parse` is called, the default `stdout` loggers will be used.

I'd like to make an rc out of this PR once merged to make sure everything is working as intended for all of the wrappers

internal/options/options.go
```diff
@@ -191,3 +209,11 @@ func GetLDOptionsFromEnv() (map[string]string, error) {
 
 	return ldOptions, nil
 }
+
+func GetDebugOptionFromEnv() (bool, error) {
+	debug := os.Getenv("debug")
```
vlymar: i think the convention is for env vars to be all uppercase. also, i'm a fan of making specific env var flags like (like `LDDEBUG`), because there's a high chance that `DEBUG` will collide with some other env var names and trigger debug behavior in some other program.

atrakh: ooh- i think this was an oversight, the variable was supposed to be `LD_DEBUG`. good catch!

PR #63, opened 2019-02-06
add version command line argument

atrakh:
`ld-find-code-refs` will now always log the current version number to stdout. If the `version` argument is provided, `ld-find-code-refs` will immediately terminate after logging the version number.

### DIFF
--- a/pkg/coderefs/coderefs.go
+++ b/pkg/coderefs/coderefs.go
@@ -20,6 +20,8 @@ import (
 // PUTing a massive json payload. These limits will likely be tweaked over
 // time. The LaunchDarkly backend will also apply limits.
 const (
+	version = "0.6.0"
+
 	minFlagKeyLen                     = 3
 	maxFileCount                      = 5000
 	maxLineCharCount                  = 500
@@ -60,6 +62,10 @@ type branch struct {
 }
 
 func Scan() {
+	log.Stdout.Printf("ld-find-code-refs version " + version)
+	if o.Version.Value() {
+		return
+	}
 	cmd, err := command.NewClient(o.Dir.Value())
 	if err != nil {
 		log.Error.Fatalf("%s", err)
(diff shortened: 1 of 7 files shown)


pkg/coderefs/coderefs.go
```diff
@@ -20,6 +20,8 @@ import (
 // PUTing a massive json payload. These limits will likely be tweaked over
 // time. The LaunchDarkly backend will also apply limits.
 const (
+	version = "0.6.0"
```
vlymar: i know we need this but its unfortunate that we have yet another thing we need to update every single time. Perhaps RELEASING.md should have a checklist of all lines of code that need updating, or a sed script that does it or something

pkg/coderefs/coderefs.go
```diff
@@ -60,6 +62,10 @@ type branch struct {
 }
 
 func Scan() {
+	log.Stdout.Printf("ld-find-code-refs version " + version)
```
vlymar: its annoying but we should probably move this up higher in the callstack. may require duplicating it to each entrypoint. but, consider:

`cmd/ld-find-code-refs/main.go`:
```
func main() {
	err, cb := o.Init()
	if err != nil {
		log.Init(false)
		log.Error.Printf("could not validate command line options: %s", err)
		cb()
		os.Exit(1)
	}
	log.Init(o.Debug.Value())
	coderefs.Scan()
}
```
Someone can't figure out how to get their params set up right, so they submit a support ticket with the output. We won't automatically see their version number.

Drawback of this is we'll need to duplicate this log statement in each entrypoint (GA, bitbucket, cmd, etc)

vlymar: counterargument is its better to just keep it in one place, and we can always ask someone to just re-run it with `-version` and submit that

atrakh: You bring up a good point. I also just realized that `--version` won't work if it's the only parameter sent over, because the program fails on the required arguments, which weren't provided.

This PR needs a little work to make --version usable

PR #64, opened 2019-02-06
add support for .ldignore files

atrakh:
Adds support for `.ldignore` files using [`go-gitignore`](https://github.com/sabhiram/go-gitignore). Since there is a new vendored dependency, the diff shows many lines of code added, but there's really only ~50 LOC changed

New behavior described here: https://github.com/launchdarkly/ld-find-code-refs/blob/d82c6fd01bcf3eb49b57723dd66cd031840f1cb8/README.md#ignoring-files-and-directories


pkg/coderefs/coderefs.go
```diff
@@ -170,21 +183,22 @@ func getFlags(ldApi ld.ApiClient) ([]string, error) {
...
 }
 
-func generateReferencesFromGrep(flags []string, grepResult [][]string, ctxLines int, exclude *regexp.Regexp) []grepResultLine {
+func generateReferencesFromGrep(flags []string, grepResult [][]string, ctxLines int, ldIgnore *ignore.GitIgnore, exclude *regexp.Regexp) []grepResultLine {
 	references := []grepResultLine{}
 
 	for _, r := range grepResult {
 		path := r[1]
-		if exclude != nil && exclude.String() != "" && exclude.MatchString(path) {
+		if (exclude != nil && exclude.String() != "" && exclude.MatchString(path)) ||
```
ashanbrown: You could write these as multiple statements without the extra parens and the broken line.   I suspect that if you ran golangci-lint on this, it would flag it.

pkg/coderefs/coderefs_test.go
```diff
@@ -6,11 +6,14 @@ import (
 	"strings"
 	"testing"
 
+	ignore "github.com/sabhiram/go-gitignore"
 	"github.com/stretchr/testify/require"
 
 	"github.com/launchdarkly/ld-find-code-refs/internal/ld"
 )
 
+var ignoreAll, _ = ignore.CompileIgnoreLines("*")
```
ashanbrown: Maybe you need a `MustCompileIgnoreLines` helper.  ;)

atrakh: I think it's fine considering it's compiling a hardcoded variable and we only use this once and only in tests. If we do end up using this more, I'll implement this feedback

PR #66, opened 2019-02-07
Simplify circle orb config, add usage examples

atrakh:
This PR implements some feedback from the Circle team about our orb:
- Add a link back to our docs in the root description
- Don't map to env
- Remove executor, inlining the `ld-find-code-refs` docker image spec and command into the job
- Add several usage examples

Other tweaks:
- Fixed a couple typos
- Added configurable `access_token` parameter, which is the `LD_ACCESS_TOKEN` by default.
- Added `debug` parameter

### DIFF
--- a/build/package/circleci/orb.yml
+++ b/build/package/circleci/orb.yml
@@ -2,7 +2,7 @@ version: 2.1
 description: "Job for finding and sending feature flag code references to LaunchDarkly. Code references documentation: https://docs.launchdarkly.com/v2.0/docs/git-code-references"
 
 examples:
-  minimal:
+  minimal_config:
     description: Minimal configuration
     usage:
       version: 2.1
@@ -47,7 +47,7 @@ examples:
                 context_lines: 3
                 exclude: 'vendor/|\.css'
 
-  :
+  standard_configuration:
     description: "A configuration with the the `repoType` set to GitHub, and the `repuUrl` set to a GitHub URL. We recommend configuring these parameters so LaunchDarkly is able to generate reference links to your source code"
     usage:
       version: 2.1
--- a/build/package/circleci/orb.yml
+++ b/build/package/circleci/orb.yml
@@ -1,61 +1,5 @@
 version: 2.1
-description: Job for finding and sending feature flag code references to LaunchDarkly
-
-executors:
-  default:
-    parameters:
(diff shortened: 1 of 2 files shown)


build/package/circleci/orb.yml
```diff
@@ -1,65 +1,77 @@
...
+      version: 2.1
+      orbs:
+        launchdarkly: launchdarkly/ld-find-code-refs@0.5.0
+      workflows:
+        main:
+          jobs:
+            - launchdarkly/find-code-references:
+                access_token: "${LD_ACCESS_TOKEN}"
+                proj_key: 'YOUR_LAUNCHDARKLY_PROJECT_KEY'
+                debug: true
```
vlymar: out of curiosity, why set debug:true in a minimal config

build/package/circleci/orb.yml
```diff
@@ -95,16 +108,30 @@ jobs:
...
+              -accessToken=<< parameters.access_token >> \
+              -projKey=<< parameters.proj_key >> \
+              -exclude=<< parameters.exclude >> \
+              -contextLines=<< parameters.context_lines >> \
+              -baseUri=<< parameters.base_uri >> \
+              -repoType=<< parameters.repo_type >> \
+              -repoUrl=<< parameters.repo_url >> \
+              -defaultBranch=<< parameters.default_branch >> \
+              -commitUrlTemplate=<< parameters.commit_url_template >> \
+              -hunkUrlTemplate=<< parameters.hunk_url_template >> \
```
vlymar: our new patch logic still works when `-hunkUrlTemplate=""` right? or do we asasume that the `-hunkUrlTemplate` is never provided at all.

build/package/circleci/orb.yml
atrakh: Given that debug logs help so much with potential configuration errors, I think having it turned on in examples will help nudge users to turn it on

build/package/circleci/orb.yml
atrakh: If the `hunk_url_template` parameter is not provided, `-hunkUrlTemplate` will resolve to an empty string, which is the same as the default. So the new(ish) logic around templates in `ld-find-code-refs` will still work as intended.

PR #67, opened 2019-02-11
log an error when no hunks were made for a reference

atrakh:
This PR adds a helper log message in case we run into an error generating hunks. This shouldn't ever happen, but if it does, this log will help us debug any potential issues.


pkg/coderefs/coderefs.go
```diff
@@ -247,6 +247,12 @@ func (g grepResultLines) makeReferenceHunksReps(projKey string, ctxLines int) []
 
 		hunks := fileGrepResults.makeHunkReps(projKey, ctxLines)
 
+		if len(hunks) == 0 {
+			log.Error.Printf("unknown error scanning for code references in %s", fileGrepResults.path)
```
ashanbrown: I'd suggest saying something like `expected code references but found none in "%s"`

PR #68, opened 2019-02-11
add precommit config

atrakh:
Adds a `pre-commit` hook config for `ld-find-code-refs`. `pre-commit` is installed via `make init`, which is already documented as a dev step in `TESTING.md`. `make lint` now runs pre-commit instead of golangci-lint directly.

Addressed a few easy golangci-lint errors, excluded all others that were a bit more intricate


internal/options/options.go
```diff
@@ -12,68 +12,67 @@ import (
...
-// Can't wait for contracts
 type Option interface {
 	name() string
 }
 
-type StringOption string
-type IntOption string
-type Int64Option string
-type BoolOption string
+type stringOption string
```
atrakh: made these unexported because we don't use them anywhere. this is an internal package, so this is not a breaking change

PR #73, opened 2019-02-12
add entrypoint to cmd docker image

atrakh:
Adds the `ld-find-code-refs` tool as an entrypoint to the command-line docker image.

So, the docker image can now run `ld-find-code-refs` directly, rather than running command against the image. ex:
```
docker run launchdarkly/ld-find-code-refs -accessToken="api-xxx" -projKey="default" -repoName="SupportService"
```


build/package/circleci/orb.yml
```diff
@@ -114,6 +114,7 @@ jobs:
         default: false
     docker:
       - image: launchdarkly/ld-find-code-refs:0.6.0
+        entrypoint: sh
```
atrakh: I spent some time tweaking the circle orb config to try to get it to place nice with the entrypoint, but was not able to do anything that clean. So, I resorted to overriding the entrypoint in the circle orb instead of coming up with a simpler config.

If anyone is familiar with how to do this sort of thing, I'd appreciate feedback on how we could make this better.

README.md
```diff
@@ -60,14 +60,14 @@ choco install ag
...
-
-<!-- TODO: update with entrypoint execution when available -->
-
-```shell
+```bash
 docker pull launchdarkly/ld-find-code-refs
-docker run -it launchdarkly/ld-find-code-refs
+docker run \
+  -v /Users/arnold/Documents/projects/launchdarkly/support-service:/repo \
+  launchdarkly/ld-find-code-refs \
```
ashanbrown: This line should not be necessary if entrypoint is set to ld-find-code-refs

atrakh: How so? `launchdarkly/ld-find-code-refs` is the image name

build/package/circleci/orb.yml
ashanbrown: This seems right to me.

PR #74, opened 2019-02-13
add `delimiters` command-line option, use strict delimiters by default

atrakh:
This PR replaces the existing word-boundary constraint on flag key matches (`\b`, or the `--word-regexp` option for `ag`), with a more-strict requirements for delimiters.

By default, single-quotes, double-quotes, and backticks are valid delimiters. These defaults are **always** applied. Users can add additional delimiters using the `delimiters` (or `d` for short) command line option. For multiple delimiters, `-d` may be supplied more than once, or can be a string with multiple characters in it. Examples of correctly configuring the letters a and b as delimiters:
`-delimiters="ab"`
`-D="ab"`
`-D="a" -D="b"`

Delimiters are single-character, non-control ASCII symbols used to detect flag keys. Any delimiter can match on either the beginning or end of a flag key. 
Example strings that will match with default delimiters for `flag`:
`"flag"`
`'flag'`
`"flag'` <-- notice the asymmetry

Matches are asymmetrical to provide support for some of the more exotic delimiter configurations, such as erlang:
`<<flag>>` (in such a case, we would provide `-delimiters="<>"`)


internal/command/command.go
```diff
@@ -81,60 +82,75 @@ func (c Client) branchName() (string, error) {
...
+func generateFlagRegex(flags []string) string {
+	flagRegexes := []string{}
+	for _, v := range flags {
+		escapedFlag := regexp.QuoteMeta(v)
+		flagRegexes = append(flagRegexes, escapedFlag)
+	}
+	return strings.Join(flagRegexes, "|")
+}
+
+func generateDelimiterRegex(delimiters options.RuneSet) (lookBehind, lookAhead string) {
```
ashanbrown: Using `options.RuneSet` here, allows configuration code to leak into this.  I'd suggest just passing a `[]Rune` here.

internal/command/command_test.go
```diff
@@ -0,0 +1,92 @@
...
+func TestGenerateFlagRegex(t *testing.T) {
+	specs := []struct {
+		name     string
+		flags    []string
+		expected string
+	}{
+		{
+			name:     "succeeds for single flag",
+			flags:    []string{"flag"},
+			expected: "flag"},
```
ashanbrown: Just a minor style thing, but I think this would be easier to read if the `}` were on the next line.

Issue #78, opened 2019-03-03
Include file types


gtevans:
Do you have any plans to have an include file types parameter rather than an ignore file types? My scenario would be to search in *.ts and *.cs files, which would be a lot easier to include these rather than exclude all others.

Thanks

atrakh:
Hi @gtevans,

For anything more advanced a simple directory or filetype exclude regex, we'd recommend specifying  a [`.ldignore`](https://github.com/launchdarkly/ld-find-code-refs#ignoring-files-and-directories) file in the root dir of your repo to specify paths for `ld-find-code-refs`. 

The negation operator (as specified by https://git-scm.com/docs/gitignore#_pattern_format) can be used to specify includes.

Example `.ldignore` to only match `.ts` and `.cs` files:
```
!*.ts
!*.cs
```

gtevans:
@atrakh I can't get adding the file types to a .ldignore file to work.

To simplify the issue I added a .ldignore file to the root of my repository and added only a '*' in the file, with the expectation that all files should be ignored and no flags would be picked up by the tool. This wasn't the case and all flags were picked up. 

Any ideas?

atrakh:
Thanks for following up- what you've described is not expected behavior.

To troubleshoot, can you please confirm that you're running the latest version of `ld-find-code-refs`? The current latest version if 1.0.0. You can find out which version you're running by executing the command line utility with the `-version` parameter. If that command line option does not exist, it's likely you're running an older version.

If you are encountering this issue on 1.0.0, some additional information about how you're executing `ld-find-code-refs` would be useful to troubleshoot, including:
- How you've installed `ld-find-code-refs` (GitHub actions, homebrew, downloaded a release, etc.)
- Which operating system `ld-find-code-refs` is executing on
- If you're executing CLI manually, it would be useful to see the command being executed

gtevans:
@atrakh I am on version 1.0.0, I downloaded the release via github, using ld-find-code-refs_1.0.0_windows_amd64, running the cli manually.

Command - ld-find-code-refs.exe -accessToken=api-*** -projKey="***" -repoName="***" -dir="***" -debug=true -repoUrl="http://***" -contextLines=5 -repoUrl="http://***" -commitUrlTemplate="http://***/${sha}"

Can you update the exe to output a debug statement if the .ldignore file is found / used?

If you want to see the output of the application can you share your email address and I can send over a screencast.

atrakh:
[...]

PR #79, opened 2019-03-12
Use filepath.Join on .ldignore path

atrakh:
This PR makes a couple changes in attempt to resolve the bug reported in https://github.com/launchdarkly/ld-find-code-refs/issues/78.

I haven't yet been able to reproduce this issue on Windows 10 in Parallels, but I did notice that there's a potential issue with filepath resolution when adding the ignore path to `ag`. Also added a debug log when `.ldignore` is found


internal/command/command.go
```diff
@@ -91,7 +91,12 @@ func (c Client) revParse(branch string) (string, error) {
 }
 
 func (c Client) SearchForFlags(flags []string, ctxLines int, delimiters []rune) ([][]string, error) {
-	args := []string{"--nogroup", "--case-sensitive", "--path-to-ignore=" + c.Workspace + "/.ldignore"}
+	args := []string{"--nogroup", "--case-sensitive"}
+	pathToIgnore := filepath.Join(c.Workspace, ".ldignore")
```
ashanbrown: Might be worth putting `.ldignore` in a constant.

PR #80, opened 2019-03-28
Delete stale branches

atrakh:
This PR adds new functionality to `ld-find-code-refs`. The scanner will now search for "stale" branches and delete the connection to that branch from LaunchDarkly.

A branch is stale if it does not exist on the Git remote of the currently selected branch. 

If the scanner fails to find or delete stale branches, it will continue with the scan (not a fatal error)

### DIFF
--- a/pkg/coderefs/coderefs.go
+++ b/pkg/coderefs/coderefs.go
@@ -92,8 +92,6 @@ func Scan() {
 		log.Error.Fatalf(err.Error())
 	}
 
-	deleteStaleBranches(ldApi, cmd, repoParams.Name)
-
 	flags, err := getFlags(ldApi)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/coderefs/coderefs.go
```diff
@@ -92,6 +92,8 @@ func Scan() {
 		log.Error.Fatalf(err.Error())
 	}
 
+	deleteStaleBranches(ldApi, cmd, repoParams.Name)
```
vlymar: An alternative to consider - don't swallow all errors inside `deleteStaleBranches`, and move it to the bottom of this function. I think swallowing all errors could lead to us missing issues down the line. If deletion is the last thing that happens, coderefs will still be posted, but users will also notice failures and report issues.

PR #81, opened 2019-04-03
Add optional support for ripgrep search

atrakh:
As discovered in #78 (and discussed further in an email thread), `ag` has a [wontfix bug](https://github.com/ggreer/the_silver_searcher/issues/611#issuecomment-75513768) that limits the ability to use the negation operator in `.*ignore`. I believe they did add some support for negation operators after the linked comment was written, but it still doesn't seem to support more complicated negation patterns, like discovering negated excludes in nested directories.

So, this PR adds a new option `-searchTool` option which allows users to specify whether `ld-find-code-refs` should use `ag` or `rg`


internal/command/command.go
```diff
@@ -13,23 +13,15 @@ import (
...
 
-/*
-grepRegex splits resulting grep lines into groups
-Group 1: File path.
-Group 2: Separator. A colon indicates a match, a hyphen indicates a context lines
-Group 3: Line number
-Group 4: Line contents
-*/
-var grepRegex = regexp.MustCompile("([^:]+)(:|-)([0-9]+)[:-](.*)")
-
```
atrakh: Moved this out of the global scope

internal/command/command.go
```diff
@@ -41,9 +33,18 @@ func NewClient(path string) (Client, error) {
...
+	_, err = exec.LookPath(string(searchTool))
 	if err != nil {
-		return client, errors.New("ag (The Silver Searcher) is a required dependency, but was not found in the system PATH")
+		switch searchTool {
+		case SilverSearcher:
+			return client, errors.New("ag (The Silver Searcher) is a required dependency, but was not found in the system PATH")
+		case Ripgrep:
+			return client, errors.New("rg (ripgrep) is a required dependency, but was not found in the system PATH")
+		default:
+			return client, fmt.Errorf("%s is not a valid search tool", searchTool)
```
atrakh: We shouldn't ever get here, but add this `default` guard just to be safe

internal/command/command.go
```diff
@@ -114,7 +158,19 @@ func (c Client) SearchForFlags(flags []string, ctxLines int, delimiters []rune)
...
+	   searchRegex splits resulting grep lines into groups
+	   Group 1: File path.
+	   Group 2: Line number.
+	   Group 3: Separator. A colon indicates a match, a hyphen indicates a context lines
+	   Group 4: Line contents.
+
+	   Example output from line search tool: /path/to/file/flags.go:2:ldClient.BoolVariation("my-flag-key", user, false)
+	   Example result: ["/path/to/file/flags.go:2:ldClient.BoolVariation(\"my-flag-key\", user, false)", "/path/to/file/flags.go", 2, ":", "ldClient.BoolVariation(\"my-flag-key\", user, false)"]
+	*/
+	searchRegex := regexp.MustCompile("(.*)[:-]([0-9]+)(:|-)(.*)")
```
atrakh: This regex moved, so the diff won't show that it did change a bit:
`([^:]+)(:|-)([0-9]+)[:-](.*)` -> `(.*)[:-]([0-9]+)(:|-)(.*)`

This change matches the content slightly differently, for a couple reasons:
- `rg` "context" lines look slight different from the ones returned from `ag` in that the separator between the filepath and line number is a hyphen instead of a colon. The matching needed to be updated to match correctly while not breaking matching on filepaths with hyphens on them
- The group matching the separator was matching the one that was always a colon (bug), when we really needed to match the one that could be a colon or a hyphen. So, the regex has been updated to match the correct item. This resulted in the groups being reordered.

pkg/coderefs/coderefs.go
```diff
@@ -178,20 +178,19 @@ func (b *branch) findReferences(cmd command.Client, flags []string, ctxLines int
 	if err != nil {
 		return grepResultLines{}, err
 	}
-
 	return generateReferencesFromGrep(flags, grepResult, ctxLines, exclude), nil
 }
 
-func generateReferencesFromGrep(flags []string, grepResult [][]string, ctxLines int, exclude *regexp.Regexp) []grepResultLine {
+func generateReferencesFromGrep(flags []string, grepResult command.SearchResults, ctxLines int, exclude *regexp.Regexp) []grepResultLine {
```
atrakh: This file still refers to "grep". We should change that to just generically be "search", but I figured there's enough changes in this PR to warrant doing this at another time.

internal/command/command.go
```diff
@@ -114,7 +158,19 @@ func (c Client) SearchForFlags(flags []string, ctxLines int, delimiters []rune)
...
+
+	searchRegexWithFilteredPath, err := regexp.Compile("(?:" + regexp.QuoteMeta(c.Workspace) + "[\\/]?)" + searchRegex.String())
```
atrakh: This regex wrapper has been changed slightly to support the windows path separator (backslash)

Issue #91, opened 2019-09-03
Unable to run against larger repo


cuzzasoft:
I'm getting the below when running this against a larger repo:

```
docker run --rm -v /repo:/repo launchdarkly/ld-find-code-refs -dir="/repo" -repoName="frontend" -projKey="cats" -accessToken="${LD_AUTH}"

ld-find-code-refs version 1.2.0
INFO: 2019/09/03 00:12:54 command.go:199: absolute directory path: /repo
INFO: 2019/09/03 00:12:54 command.go:56: git branch: master
INFO: 2019/09/03 00:13:10 coderefs.go:215: finding code references with delimiters: ["'`]
ERROR: 2019/09/03 00:13:10 coderefs.go:130: error searching for flag key references: ERR: Bad regex! pcre_compile() failed at position 108921: regular expression is too large
If you meant to search for a literal string, run ag with -Q
```

Happens with both Docker image and also running it on Mac local machine. We have around 2k flags in the project I'm running it against.

Is there any way we can increase the regex limit, or paginate the execution of the app?

cuzzasoft:
I 'fixed' this with the below:

```
diff --git a/.gitignore b/.gitignore
index e5fc37e..18cd1c2 100644
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ build/package/bitbucket-pipelines/ld-find-code-refs-bitbucket-pipeline
 build/package/cmd/ld-find-code-refs
 /dist/
 /bin/
+.idea/
diff --git a/pkg/coderefs/coderefs.go b/pkg/coderefs/coderefs.go
index 9f49c19..f794045 100644
--- a/pkg/coderefs/coderefs.go
+++ b/pkg/coderefs/coderefs.go
@@ -213,12 +213,26 @@ func getFlags(ldApi ld.ApiClient) ([]string, error) {
 func (b *branch) findReferences(cmd command.Client, flags []string, ctxLines int, exclude *regexp.Regexp) (grepResultLines, error) {
 	delims := o.Delimiters.Value()
 	log.Info.Printf("finding code references with delimiters: %s", delims.String())
-	grepResult, err := cmd.SearchForFlags(flags, ctxLines, delims)
-	if err != nil {
-		return grepResultLines{}, err
+
+	var pageSize = 50
+	var grepResults [][]string
+	log.Info.Printf("paginating %d flags with a pagesize of %d", len(flags), pageSize)
[...]

PR #92, opened 2019-09-04
Cs/add pagination support for mammoth repos

cuzzasoft:
This adds support for pagination as suggested in #91. We have a repo with ~ 2.4k feature flags and I ended up bumping pagination to 500 to make the job go a bit faster. Based on the conversation around sorting I tried to sort grepLines by path and then lineNum, however I'm new to Golang and not so familiar with the algorithm / not much time to spend on it so hopefully you can fix it up or let me know what to fix if it's bung.

Thanks :)

### DIFF
--- a/pkg/coderefs/coderefs.go
+++ b/pkg/coderefs/coderefs.go
@@ -213,12 +213,26 @@ func getFlags(ldApi ld.ApiClient) ([]string, error) {
 func (b *branch) findReferences(cmd command.Client, flags []string, ctxLines int, exclude *regexp.Regexp) (grepResultLines, error) {
 	delims := o.Delimiters.Value()
 	log.Info.Printf("finding code references with delimiters: %s", delims.String())
-	grepResult, err := cmd.SearchForFlags(flags, ctxLines, delims)
-	if err != nil {
-		return grepResultLines{}, err
+
+	var pageSize = 50
+	var grepResults [][]string
+	log.Info.Printf("paginating %d flags with a pagesize of %d", len(flags), pageSize)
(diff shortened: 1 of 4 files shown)


pkg/coderefs/coderefs.go
```diff
@@ -158,6 +160,19 @@ func Scan() {
 	}
 }
 
+// Sorts grep results by Path and then line number
+func sortGrepResults(lines grepResultLines) {
```
cuzzasoft: I put it into a function solely for testing

bwoskow-ld:
Thanks for the pull request, @cuzzasoft! As @atrakh mentioned on #91, we'll likely want to make additional changes before merging this in, but this is a really helpful start. We'll keep you updated on this pull request.

pkg/coderefs/coderefs.go
```diff
@@ -158,6 +160,19 @@ func Scan() {
...
+	sort.Slice(lines, func(i, j int) bool {
```
atrakh: Since we already have the `grepResultLines` type, we can implement [`sort.Interface`](https://golang.org/pkg/sort/#Interface) on `grepResultsLines` and then call `sort.Sort(linesToSort)`, which would replace the need for a separate function for sorting results

cuzzasoft:
Thanks for the feedback, I've been meaning to get to this but been a bit slammed this week. I should be able to fix this up Monday next week with the requested changes :)

atrakh:
I'm closing this PR as we've merged #94 (based off the changes in this pull request) into a release candidate branch, which we hope to release in the coming days. Thanks again for your help!

I'll follow up again on #91 once we release.

PR #93, opened 2019-09-10
Always apply delims

atrakh:
We currently match against delimiters when performing the coderefs search against a repository, but don't validate the delimiters later on when we pick out specific flag keys in a given line. This PR makes sure we always match against delimiters when comparing flag keys against lines.

### DIFF
--- a/pkg/coderefs/coderefs.go
+++ b/pkg/coderefs/coderefs.go
@@ -2,6 +2,7 @@ package coderefs
 
 import (
 	"container/list"
+	"fmt"
 	"os"
 	"regexp"
 	"strconv"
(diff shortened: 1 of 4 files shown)


pkg/coderefs/coderefs.go
```diff
@@ -238,7 +239,7 @@ func generateReferencesFromGrep(flags []string, grepResult [][]string, ctxLines
 		}
 		ref := grepResultLine{Path: path, LineNum: lineNum}
 		if contextContainsFlagKey {
-			ref.FlagKeys = findReferencedFlags(lineText, flags)
+			ref.FlagKeys = findReferencedFlags(lineText, flags, delims)
```
vlymar: ah, so `contextContainsFlagKey` tells us that this line has a flag reference, but in `findReferencedFlags` we'll loop over all flags and return the first flag that matches, regardless of delims.

pkg/coderefs/coderefs_test.go
```diff
@@ -37,90 +49,101 @@ func Test_generateReferencesFromGrep(t *testing.T) {
...
-				{Path: "flags.txt", LineNum: 12, LineText: "someFlag", FlagKeys: []string{"someFlag"}},
+				testWant,
 				{Path: "flags.txt", LineNum: 13, LineText: "not a flag key line"},
-				{Path: "flags.txt", LineNum: 14, LineText: "anotherFlag", FlagKeys: []string{"anotherFlag"}},
+				{Path: "flags.txt", LineNum: 14, LineText: delimit(testFlagKey2, `"`), FlagKeys: []string{testFlagKey2}},
 				{Path: "flags.txt", LineNum: 15, LineText: "not a flag key line"},
 			},
 		},
+		{
+			name:       "does not match substring flag key",
```
vlymar: 👍 for good tests

PR #96, opened 2019-09-12
add fatal logger message

atrakh:
Adds a call to action to file a github issue when a user encounters an unrecoverable fatal error

### DIFF
--- a/pkg/coderefs/coderefs.go
+++ b/pkg/coderefs/coderefs.go
@@ -60,7 +60,6 @@ type branch struct {
 }
 
 func Scan() {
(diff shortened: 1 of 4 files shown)


pkg/coderefs/coderefs.go
```diff
@@ -60,10 +60,11 @@ type branch struct {
 }
 
 func Scan() {
+	log.Fatal.Fatalf("blah blah blah")
```
vlymar:
```suggestion
```

atrakh: 🤦‍♂

PR #97, opened 2019-09-13
add options for dry run and export to csv

atrakh:
- Added a `--outDir` option to the CLI. If provided, code references will be written to a csv file in `outDir`.
- Added a `--dryRun` option to the CLI. If provided, `ld-find-code-refs` will scan for code references without sending them to LaunchDarkly. May be used in conjunction with `--outDir` to output code references data to a csv file instead of sending data to LaunchDarkly.


atrakh:
This requires changes to the README, which I'll do in `rc-1.3.0`

internal/command/command.go
```diff
@@ -190,36 +193,3 @@ func generateSearchPattern(flags []string, delimiters []rune, padPattern bool) s
 	// example: (?<=[\"'\`])(flag1|flag2|flag3)(?=[\"'\`])"
 	return lookBehind + "(" + flagRegex + ")" + lookAhead
 }
-
-func normalizeAndValidatePath(path string) (string, error) {
```
atrakh: Factored this out into a new package since we're using it in multiple locations now

internal/ld/ld.go
```diff
@@ -389,11 +392,59 @@ func (b BranchRep) TotalHunkCount() int {
...
+	}
+
+	// sort csv by flag key
+	sort.Slice(records, func(i, j int) bool {
+		// sort by flagKey -> path -> startingLineNumber
+		for k := 0; k < 3; k++ {
+			if records[i][k] != records[j][k] {
+				return records[i][k] < records[j][k]
+			}
+		}
```
atrakh: This loop is not exactly my favorite way to sort, but I think it's safe to assume the ordering of these values are invariant

internal/ld/ld.go
```diff
@@ -389,11 +392,59 @@ func (b BranchRep) TotalHunkCount() int {
 	return count
 }
 
+func (b BranchRep) WriteToCSV(outDir, projKey, repo, sha string) (path string, err error) {
+	tag := b.Name
+	if len(sha) >= 7 {
+		tag = sha[:7]
+	}
```
atrakh: Try to create a filename with a shortened sha, but if the sha is too short for some unexpected reason, use the branch name instead

atrakh:
This PR may change slightly as I merge code in from the release candidate branch: https://github.com/launchdarkly/ld-find-code-refs/tree/rc-1.3.0

README.md
```diff
@@ -188,10 +188,12 @@ Although these arguments are optional, a (\*) indicates a recommended parameter
...
+| `dryRun`            | If enabled, `ld-find-code-refs` will scan for code references without sending them to LaunchDarkly. May be used in conjunction with `outDir` to output code references data to a csv file instead of sending data to LaunchDarkly.                                                                                                                                                                                                                                       | `false`                        |
+| `outDir`            | Path to an existing directory. If provided, code references will be written to a csv file in the `outDir`                                                                                                                                                                                                                                                                                                                                                                |
```
bwoskow-ld: You should mention what the filename would be in case they specify a directory already containing many csv files.

internal/ld/ld.go
```diff
@@ -389,11 +392,59 @@ func (b BranchRep) TotalHunkCount() int {
...
+
+	absPath, err := validation.NormalizeAndValidatePath(outDir)
+	if err != nil {
+		return "", fmt.Errorf("invalid outDir '%s': %s", outDir, err)
+	}
+	path = filepath.Join(absPath, fmt.Sprintf("coderefs_%s_%s_%s.csv", projKey, repo, tag))
+
+	f, err := os.Create(path)
```
bwoskow-ld: Will this create the parent directory if it doesn't already exist?

bwoskow-ld: Nevermind; now I came across https://github.com/launchdarkly/ld-find-code-refs/pull/97/files#diff-bd40f10a5bef7a0945987430ac7762bdR21

atrakh: Good find-- yes, the validation code is called at the start of execution in `options.go`, ensuring that the directory does not exist. I'll add a note in the readme that the directory must already exist.

internal/ld/ld.go
ldhenry: very nit: but I generally tend to prefer putting the unhappy path cases in the if statement and let the happy path cases be the default.

internal/validation/validation.go
```diff
@@ -0,0 +1,39 @@
...
+	if !exists {
+		return "", fmt.Errorf("directory does not exist: %s", absPath)
+	}
+
+	return absPath, nil
+}
+
+func dirExists(path string) (bool, error) {
+	fileInfo, err := os.Stat(path)
+
```
ldhenry: nit: this function looks a little too spaced out visually. I'd probably strip the empty lines except after if blocks to match the convention you used in `NormalizeAndValidatePath`

Issue #103, opened 2019-10-29
ld-find-code-refs fails closed on API failures


garthk:
Build broken by `ld-find-code-refs` 1.2.0–1.3.1:
```
$ .bin/ld-find-code-refs -accessToken=$LAUNCH_DARKLY_ACCESS_TOKEN -branch $GIT_BRANCH -contextLines=2 -dir=. -projKey=default -repoName=$REPO -repoType=github -repoUrl=https://github.com/$ORG/$REPO
ld-find-code-refs version 1.3.1
INFO: 2019/10/29 06:48:47 coderefs.go:63: absolute directory path: /home/ci/$REPO
INFO: 2019/10/29 06:48:47 git.go:39: git branch: fix/xxx-yyy-zzz-#999999
INFO: 2019/10/29 06:48:48 search.go:87: finding code references with delimiters: ["'`]
INFO: 2019/10/29 06:48:48 coderefs.go:170: sending 7 code references across 2 flags and 3 files to LaunchDarkly for project: default
FATAL Please file an issue at https://github.com/launchdarkly/ld-find-code-refs: 2019/10/29 06:48:48 coderefs.go:183: error sending code references to LaunchDarkly: invalid_request, field 'name' must be empty or equal to path parameter 'branch'
```

atrakh:
Hi @garthk. Thank you for bringing this to our attention. We've identified a regression in our application causing the branch-updating api route to incorrectly return an `invalid_request` status.

We're working on a fix for this now-- I'll follow-up on this issue once the fix has been deployed. Since this is an issue in our application, existing installations of `ld-find-code-refs` will start working again once the fix has been released, without requiring an upgrade.

garthk:
@atrakh, assuming there'll always be the possibility of back end failure, could you arrange for ld-find-code-refs to \[be able to] exit with status 0 on failures behind the API server? If the API server is in good enough shape to reply with a well formed error other than an authentication or access control failure, perhaps we don't need our build broken. Not having this commit's lines updated is a mild inconvenience. Breaking our build wasn't mild.

atrakh:
Following up on this- a fix has been released and deployed. You should no longer see this error.

Thanks for the feedback, having a build break due to an API failure on our end is certainly not pleasant. Exiting with status 0 could be one way of making sure your workflows don't break in these cases. I'm hesitant to do that by default, as it could result in silent failures in this cass. We'll think through this for our next release `ld-find-code-refs` release.

[...]

PR #107, opened 2020-03-10
add option to control exit code on api failure

atrakh:
re: #103 

Adds a boolean option, `ignoreServiceErrors`. Controls the exit code returned by `ld-find-code-refs` upon any API error (1 if true, 0 if false). Set to `false` by default.


.circleci/config.yml
```diff
@@ -79,7 +79,7 @@ jobs:
       - setup_remote_docker
       - run: sudo apt-get update
       - run: sudo apt-get install rpm
-      - run: curl -sL https://git.io/goreleaser | bash -s -- --rm-dist --skip-publish --skip-validate --release-notes <(make echo-release-notes)
+      - run: curl -sL https://git.io/goreleaser | VERSION=v0.118.1 bash -s -- --rm-dist --skip-publish --skip-validate --release-notes <(make echo-release-notes)
```
atrakh: latest version of goreleaser was breaking the build, so I locked the version to the last one that worked.

atrakh:
Right, we definitely need to handle auth failures. I can go either way on rate limiting - if you're building fast enough to hit the rate limit, the problem should resolve itself eventually.

I like the idea of making the exit code configurable. Will follow up with that change, although I think we can iterate on the name.

atrakh:
Tying in offline conversations: we decided on `ignoreServiceErrors`, with a default to `false`, maintaining existing behavior for the default configuration.

PR #113, opened 2020-04-23
support glob patterns for filePattern alias matching

atrakh:
Replaces the `path` options with `paths`, an array of globs. This allows for more succinct declaration of alias search.

### DIFF
--- a/internal/options/alias.go
+++ b/internal/options/alias.go
@@ -9,12 +9,15 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 	"time"
 
 	"github.com/iancoleman/strcase"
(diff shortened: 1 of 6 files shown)


internal/options/alias.go
```diff
@@ -186,6 +192,62 @@ func (a *Alias) IsValid() error {
...
+		if !validation.FileExists(path) {
+			return fmt.Errorf("filePattern '%s': could not find file at path '%s'", aliasId, path)
+		}
+		/* #nosec */
+		data, err := ioutil.ReadFile(path)
+		if err != nil {
+			return fmt.Errorf("filePattern '%s': could not process file at path '%s': %v", aliasId, path, err)
+		}
+		a.AllFileContents = append(a.AllFileContents, data...)
+	}
```
atrakh, line 247: this is the new stuff - everything else has really just been moved around

PR #114, opened 2020-04-23
rename coderefs.yaml to config.yaml

atrakh:
This frees up the `.launchdarkly` directory namespace for potential future usage.

### DIFF
--- a/internal/options/alias.go
+++ b/internal/options/alias.go
@@ -201,9 +201,12 @@ func (o *YamlOptions) IsValid() error {
 }
 
 func Yaml() (*YamlOptions, error) {
-	pathToYaml := filepath.Join(Dir.Value(), ".launchdarkly/config.yaml")
+	pathToYaml := filepath.Join(Dir.Value(), ".launchdarkly/coderefs.yaml")
 	if !validation.FileExists(pathToYaml) {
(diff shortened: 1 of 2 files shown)


internal/options/alias.go
```diff
@@ -201,9 +201,12 @@ func (o *YamlOptions) IsValid() error {
 }
 
 func Yaml() (*YamlOptions, error) {
-	pathToYaml := filepath.Join(Dir.Value(), ".launchdarkly/config.yaml")
+	pathToYaml := filepath.Join(Dir.Value(), ".launchdarkly/coderefs.yaml")
 	if !validation.FileExists(pathToYaml) {
-		return nil, nil
+		pathToYaml = filepath.Join(Dir.Value(), ".launchdarkly/coderefs.yml")
+		if !validation.FileExists(pathToYaml) {
```
annabkr, line 207: Curious about the logic here in repeating both the assignment and validation check. Would love to understand.

atrakh: This duplication is here to support both common YAML file extensions `.yaml` and `.yml`. I don't mind the duplication for now because I'd like to eventually migrate this configuration to use [viper](https://github.com/spf13/viper), which handles both extensions automatically

PR #116, opened 2020-04-27
check for yaml configuration when initializing cli options

atrakh:
My last change in https://github.com/launchdarkly/ld-find-code-refs/pull/112 inadvertently broke configuration via CLI. For now, let's duplicate the yaml read logic in both `Init` and `Populate`-- moving the configuration format to viper later on will make this code easier to read and less prone to error.

### DIFF
--- a/internal/options/options.go
+++ b/internal/options/options.go
@@ -247,6 +247,14 @@ func Init() (err error, errCb func()) {
 		}
 	}
 
+	yamlOptions, err := Yaml()
+	if err != nil {
+		return nil, nil
+	}
+	if yamlOptions != nil {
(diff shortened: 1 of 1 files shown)


internal/options/options.go
```diff
@@ -247,6 +247,14 @@ func Init() (err error, errCb func()) {
 		}
 	}
 
+	yamlOptions, err := Yaml()
+	if err != nil {
+		return nil, nil
```
ld-kyee, line 252: Are there no other types of errors you would want to return like if there were parsing errors etc?

atrakh: Ah whoops, that's a typo - we should return the error

PR #118, opened 2020-04-28
Make alias config type field case insensitive

atrakh:
Includes some readme improvements. Code changes do not break 1.5.0-beta2 functionality, so these changes will be included by overwriting 1.5.0-beta2.

### DIFF
--- a/internal/version/version.go
+++ b/internal/version/version.go
@@ -1,3 +1,3 @@
 package version
 
-const Version = "1.5.0-beta1"
+const Version = "1.5.0-beta2"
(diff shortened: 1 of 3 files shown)


internal/version/version.go
```diff
@@ -1,3 +1,3 @@
 package version
 
-const Version = "1.5.0-beta1"
+const Version = "1.5.0-beta2"
```
atrakh, line 3: 1.5.0-beta2 is already released, but I missed updating the version string in that release.

PR #120, opened 2020-05-08
Use cobra for configuration

atrakh:
This PR migrates the command line tool to use [cobra](https://github.com/spf13/cobra), [viper](https://github.com/spf13/viper), and [pflag](https://github.com/spf13/pflag) for configuration. These changes help make the tool easier to configure, with more user-friendly help messages.

[This PR](https://github.com/launchdarkly/ld-find-code-refs/pull/121) includes all the documentation updates related to these changes for this repository. The docs updates may help answer questions related to major breaking changes that come up in this PR.

In addition, this allows us to easily support configuring all options as environment variables, as well as most options in the `.launchdarkly/coderefs.yaml` file introduced with alias support.

This is a pretty large refactor, and unfortunately, there wasn't a great way to break these changes out into smaller pieces. I set the base branch to `2.x` instead of master because I'd like to introduce some other changes to include with the 2.0.0 release.


internal/log/log.go
```diff
@@ -12,8 +12,6 @@ var (
 	Info    *log.Logger
 	Warning *log.Logger
 	Error   *log.Logger
-	Fatal   *log.Logger
-	Stdout  *log.Logger
```
atrakh, line 16: I decided to remove some extra loggers as part of this PR, so there will be some seemingly random log lines update in the diff

internal/options/alias.go
```diff
@@ -200,87 +125,3 @@ func (a *Alias) IsValid() error {
 
 	return nil
 }
-
-func (a *Alias) ProcessFileContent(idx int) error {
```
atrakh, line 204: Moved to the `coderefs` package since this is not "option" code

internal/options/options.go
```diff
@@ -1,337 +1,180 @@
 package options
```
atrakh, line 1: This file was essentially rewritten.  I would recommend reviewing it as if it were a completely new file.

pkg/coderefs/alias.go
```diff
@@ -1,15 +1,33 @@
...
+	"time"
+
+	"github.com/iancoleman/strcase"
 	"github.com/launchdarkly/ld-find-code-refs/internal/helpers"
 	"github.com/launchdarkly/ld-find-code-refs/internal/options"
+	"github.com/launchdarkly/ld-find-code-refs/internal/validation"
 )
 
-func generateAliases(flags []string, aliases []options.Alias) (map[string][]string, error) {
+func generateAliases(flags []string, aliases []options.Alias, dir string) (map[string][]string, error) {
```
atrakh, line 21: This function is a port of the `alias.Generate` function that used to be in the `options` package. The main difference is that the contents of files is no longer stored on the `Alias` itself.

pkg/coderefs/coderefs.go
```diff
@@ -48,13 +47,18 @@ type fileSearchResults struct {
...
 	Head             string
-	UpdateSequenceId *int64
+	UpdateSequenceId *int
 	SyncTime         int64
 	SearchResults    searchResultLines
 }
 
-func Scan() {
-	dir := o.Dir.Value()
+func Scan(opts options.Options) {
```
atrakh, line 55: passing options gets rid of global state and sets us up to make `ld-find-code-refs` usable as a go package in the future.

pkg/coderefs/coderefs.go
```diff
@@ -258,38 +268,13 @@ func getFlags(ldApi ld.ApiClient) ([]string, error) {
 	return flags, nil
 }
 
-func generateReferences(aliases map[string][]string, searchResult [][]string, ctxLines int, delims string, exclude *regexp.Regexp) []searchResultLine {
```
atrakh, line 261: Moved to `search.go` because this function is only called from that file.

internal/command/command.go
```diff
@@ -42,12 +42,12 @@ func FlagKeyCost(key string) int {
 	return len(key) + strings.Count(key, ".")
 }
 
-func DelimCost(delims []rune) int {
+func DelimCost(delims []byte) int {
```
atrakh, line 45: When migrating to `cobra`, it made more sense to pass delimiters around as bytes instead of runes

build/package/bitbucket-pipelines/bitbucket-pipelines.go
```diff
@@ -1,49 +1,34 @@
 package main
```
atrakh, line 1: All of the changes in `build/package` are changes to wrappers to be compatible with the migration to cobra. I'd recommend looking at these last

internal/version/version.go
```diff
@@ -1,3 +1,3 @@
 package version
 
-const Version = "1.5.0-beta3"
+const Version = "2.0.0-beta1"
```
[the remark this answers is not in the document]
atrakh: Our open source tools typically follow the [semantic versioning](https://semver.org/) convention. This change counts as a major version change because it may break existing implementations / usages of `ld-find-code-refs`

internal/command/command.go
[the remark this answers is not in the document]
atrakh: Might be just a style I've picked up, but I would say it references the generic "cost" of an algorithm https://en.wikipedia.org/wiki/Analysis_of_algorithms#Cost_models

In this case, the cost of adding a delimiter to the regular expression is 2 characters (the delimiter before and after each string)

PR #123, opened 2020-05-13
Refactor search code into a separate package

atrakh:
This PR refactors the interface responsible for executing and parsing searches, `command.Searcher` -> `search.SearchClient`, to be more flexible for search tools that may return data in different formats. It also includes moving search related code into a separate package to reduce the amount of code in the `coderefs` package.

There are no functional changes in this PR, but it does put us in a better place for the following in the future:
- having `ld-find-code-refs` be available as a go package
- creating more `SearchClients` for better performance, output maintainability (ripgrep supports json output), and cross-platform support


internal/search/ag.go
```diff
@@ -0,0 +1,130 @@
...
+	if runtime.GOOS == windows {
+		output = fromWindows1252(output)
+	}
+
+	ret := searchRegexWithFilteredPath.FindAllStringSubmatch(output, -1)
+
+	return c.generateReferences(aliases, ret, ctxLines, string(delimiters)), err
+}
+
+func (c *AgClient) generateReferences(aliases map[string][]string, searchResult [][]string, ctxLines int, delims string) []SearchResultLine {
```
atrakh, line 84: These methods are now a property of `AgClient` because this code is pretty particular to the way `ag` outputs search results

internal/search/ag.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	output := string(out)
+	if runtime.GOOS == windows {
+		output = fromWindows1252(output)
+	}
+
+	ret := searchRegexWithFilteredPath.FindAllStringSubmatch(output, -1)
+
+	return c.generateReferences(aliases, ret, ctxLines, string(delimiters)), err
```
atrakh, line 81: Minor change that shouldn't change output - this method is now called to convert raw output lines to `SearchResultLines` for each (potentially paginated) search output. This is done to have the `SearchClient` interface conform to a standard interface returning `SearchResultLines`

atrakh:
putting this on ice while i explore a native search implementation

PR #130, opened 2020-06-29
update build commands to reduce binary size

atrakh:
reduces binary size by ~20% by omitting debug information from builds

ref: https://blog.filippo.io/shrink-your-go-binaries-with-this-one-weird-trick/

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,8 @@ test: lint
 lint:
 	pre-commit run -a --verbose golangci-lint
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -11,20 +11,22 @@ test: lint
 lint:
 	pre-commit run -a --verbose golangci-lint
 
+BUILD_FLAGS = -ldflags="-s -w"
```
ld-kyee, line 14: perhaps add a comment with what this does would be good or a link to the article

PR #131, opened 2020-06-29
Search commit history for last removed time

atrakh:
Adds support for the new code reference extinctions API


pkg/coderefs/coderefs.go
```diff
@@ -48,15 +48,7 @@ func Scan(opts options.Options) {
...
 	projKey := opts.ProjKey
-
-	// Check for potential sdk keys or access tokens provided as the project key
-	if len(projKey) > maxProjKeyLength {
-		if strings.HasPrefix(projKey, "sdk-") {
-			log.Warning.Printf("provided projKey (%s) appears to be a LaunchDarkly SDK key", "sdk-xxxx")
-		} else if strings.HasPrefix(projKey, "api-") {
-			log.Warning.Printf("provided projKey (%s) appears to be a LaunchDarkly API access token", "api-xxxx")
-		}
-	}
```
atrakh, line 59: refactored this into a new function below

pkg/coderefs/coderefs.go
```diff
@@ -83,11 +75,6 @@ func Scan(opts options.Options) {
 		fatalServiceError(fmt.Errorf("could not retrieve flag keys from LaunchDarkly: %w", err), ignoreServiceErrors)
 	}
 
-	if len(flags) == 0 {
-		log.Info.Printf("no flag keys found for project: %s, exiting early", projKey)
-		os.Exit(0)
-	}
```
atrakh, line 89: this check was extraneous - remove it to reduce the number of exit paths in this function

internal/options/flags.go
```diff
@@ -89,6 +89,13 @@ links to the repository for each code reference.`,
 		defaultValue: false,
 		usage: `If enabled, the scanner will terminate with exit code 0 when the
 LaunchDarkly API is unreachable or returns an unexpected response.`,
+	},
+	{
+		name:         "lookback",
+		short:        "l",
+		defaultValue: 10,
```
atrakh, line 96: This default added about 2 seconds to the overall execution time against one of our large repos on my local machine. As we test the preview version, we may want to tweak the default value if it turns out to be slow.

internal/ld/ld.go
```diff
@@ -500,15 +528,23 @@ func (t tableData) Swap(i, j int) {
 
 const maxFlagKeysDisplayed = 50
 
-func (b BranchRep) PrintReferenceCountTable() {
-	data := tableData{}
+func (b BranchRep) CountByFlag(flags []string) map[string]int64 {
```
atrakh, line 531: Factored a bit of the `PrintReferenceCountTable` out into another method so it can be used for other purposes (i.e. checking if a flag has no references for extinctions)

internal/git/git.go
```diff
@@ -95,3 +102,81 @@ func (c *Client) RemoteBranches() (map[string]bool, error) {
...
 	return ret, nil
 }
+
+type CommitData struct {
+	commit *object.Commit
+	tree   *object.Tree
+}
+
+// FindExtinctions searches commit history for flags that had references removed recently
+func (c Client) FindExtinctions(projKey string, flags []string, delimiters string, lookback int) ([]ld.ExtinctionRep, error) {
```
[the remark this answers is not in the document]
atrakh: Yeah, this is only going to work with Git at the moment. Without first class support for other VCS, we can only really do this with Git

PR #132, opened 2020-07-01
support pull_request github actions

atrakh:
The `GITHUB_REF` env variable is set to the PR number instead of the branch when the action is invoked on pull request events, so add a conditional to use info from the event payload in that case.

Fixes https://github.com/launchdarkly/find-code-references/issues/7


build/package/github-actions/github-actions.go
```diff
@@ -28,21 +29,26 @@ func main() {
...
 	}
+	ghBranch, err := parseBranch(os.Getenv("GITHUB_REF"))
+	if err != nil {
+		// The GITHUB_REF wasn't valid, so check if it's a pull request and use the pull request ref instead
+		if event.Pull != nil {
+			ghBranch = event.Pull.Head.Ref
+		} else {
+			log.Error.Fatalf("error parsing GITHUB_REF: %s", err)
+		}
+	}
```
ld-kyee, line 44: could we add a spec for this?

PR #136, opened 2020-09-30
Don't process symlink contents

d3d-z7n:
Symlinks may be present in a repository, but typically should not be
processed as they may point to non-existent files (depending on the run
context) or other files in the same repository (which will be processed
independently).

### DIFF
--- a/internal/search/files.go
+++ b/internal/search/files.go
@@ -85,7 +85,7 @@ func readFiles(ctx context.Context, files chan<- file, workspace string) error {
 				return filepath.SkipDir
 			}
 			return nil
-		} else if isDir {
+		} else if isDir || info.Mode()&os.ModeSymlink == os.ModeSymlink {
 			return nil
 		}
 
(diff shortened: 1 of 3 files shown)


d3d-z7n:
Heyo, was running into this error when trying to process one of our repos.
```
ERROR: 2020/09/30 15:03:39 coderefs.go:103: error searching for flag key references: file does not exist
```
Turns out, it had a symlink in the repo pointing to a file that would only be present on the developer's machine.

In general, couldn't think of a situation where the symlink destination *should* be processed, so removed it, here.

Disclaimer - I don't know much Go, but this seemed straightforward.

atrakh:
Hi @d3d-z7n,

Thank you for your contribution, this approach seems reasonable.

I'm the maintainer for this project, but am currently out of the office. I'll be able to test & release this change early next week. Are you able to move forward with your local build until then? 

Cheers,
Arnold

d3d-z7n:
Yep, I'm all set with a custom build. Thanks!

atrakh:
LGTM! I'm going to merge this to the upstream repo and make a small change to generalize this such that only regular files are scanned

Thanks again :)

atrakh (approved):

atrakh:
Release in 2.0.1

PR #142, opened 2021-03-09
Move packages around to use as library

InTheCloudDan:
* Move `coderefs` out of `pkg` directory which is legacy and remove `pkg`
* `alias` and `search` moved from Internal
* Make `generateAliases` a public function

There is additional work that could be done to decouple code references further but this is a first step.

For my use case search may not need to be moved but it already has public methods.

If the change proposed looks reasonable then the following is a TODO:

- [x] Add documentation around using as a library.


vlymar:
Hey @dbolson, thanks for the heavy lifting here!

> Move coderefs out of pkg directory which is legacy and remove pkg

Can you share more about `pkg` being legacy? I believe we followed the patterns in https://github.com/golang-standards/project-layout for this project but could see that being outdated now.

InTheCloudDan:
@vlymar it looks like this issue is still open on it: https://github.com/golang-standards/project-layout/issues/10

They do reference this tweet from Brad Fitzpatrick though about it being legacy: https://twitter.com/bradfitz/status/1039512487538970624

vlymar (approved):
Sorry this review took so long! Thanks for linking that github issues thread, interesting stuff.

yusinto (approved):
Apologies for the slowness, I have no excuses. Looks good to me, thank you for doing this.

PR #145, opened 2021-05-06
Update release readme

yusinto:
This updates the release readme with steps to update the github action `find-code-refs` repo and also includes the release checklist from confluence.

### DIFF
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -9,17 +9,24 @@ This project uses [goreleaser](https://goreleaser.com/) to generate github relea
 
 Make sure you update the changelog before generating a release.
 
-Once a release has been completed, update the [BitBucket pipelines](https://bitbucket.org/launchdarkly/ld-find-code-refs-pipe) repo with the new version number, and push a tag containing the version number along with your commit. Example release commit: https://bitbucket.org/launchdarkly/ld-find-code-refs-pipe/commits/0b1e920c7322cd495f4fc1a09d339342d32606e4
+Once a release has been completed:
+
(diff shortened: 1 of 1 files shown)


docs/RELEASING.md
```diff
@@ -9,18 +9,25 @@ This project uses [goreleaser](https://goreleaser.com/) to generate github relea
...
 
 ## CircleCI Orb Registry
 
 To publish the CircleCI Orb to the Orb registry, you'll need the [CircleCI CLI](https://circleci.com/docs/2.0/local-cli/) installed, and will need to run `circleci setup` and authenticate with a circle token that has github owner status.
 
 Run `make publish-dev-circle-orb TAG=$VERSION` or `make-publish-release-circle-orb TAG=$VERSION` to publish the orb to the orb registry, where `$VERSION` is the version you want to release. Running `publish-dev-circle-orb` will publish a development-tagged (e.g. `dev:0.0.1`) orb, which can be overwritten, and `publish-release-circle-orb` will publish a release version of the orb, which is immutable. Both dev and release orbs are open to the public, but development orbs are not visible in the list of registered orbs on Circle's [website](https://circleci.com/orbs/registry/?showAll=true).
 
+## Release checklist
+Finally, follow the comprehensive checklist [here](https://launchdarkly.atlassian.net/wiki/spaces/ENG/pages/938214176/ld-find-code-refs+release+checklist) to make sure you have completed all the necessary steps.
```
bwoskow-ld, line 29: This release markdown document is intended for an internal audience (no third-party developer will be releasing) and yet it is in our public repository. This is particularly awkward for cases like this where you link to an internal confluence page which non-DarkLaunchers cannot access.

Should we move the contents of this document onto our Confluence instance rather than maintaining it here? It seems so.

vlymar (changes requested):
I agree with Ben that we should avoid linking to internal pages from public repos. So we should either have the whole release checklist documented publicly, or move this whole file into your release doc

PR #147, opened 2021-05-14
fix path handling for glob patterns

InTheCloudDan:
* add test coverage for globs

### DIFF
--- a/coderefs/alias_test.go
+++ b/coderefs/alias_test.go
@@ -68,7 +68,7 @@ func Test_GenerateAliases(t *testing.T) {
 			name:  "file pattern",
 			flags: slice(testFlagKey),
 			aliases: []o.Alias{
-				filePattern(testFlagKey),
+				fileExactPattern(testFlagKey),
 			},
 			want: map[string][]string{testFlagKey: slice("SOME_FLAG")},
 		},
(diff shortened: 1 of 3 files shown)


coderefs/alias_test.go
```diff
@@ -71,6 +72,14 @@ func Test_GenerateAliases(t *testing.T) {
...
+			name:  "file wildcard pattern",
+			flags: slice(testFlagKey, testWildFlagKey),
+			aliases: []o.Alias{
+				fileWildPattern(testFlagKey),
+			},
+			want: map[string][]string{testWildFlagKey: slice("WILD_FLAG"), testFlagKey: slice("SOME_FLAG")},
+		},
 		// TODO
 		// {
 		// 	name:    "command",
```
yusinto, line 85: Is this TODO still valid or safe to remove?

coderefs/alias_test.go
```diff
@@ -71,6 +72,14 @@ func Test_GenerateAliases(t *testing.T) {
 			},
 			want: map[string][]string{testFlagKey: slice("SOME_FLAG")},
 		},
+		{
+			name:  "file wildcard pattern",
```
yusinto, line 76: Nit: I wonder if the previous test `file pattern` should be renamed to `exact filename pattern` to distinguish it from the wildcard test?

yusinto (approved):

coderefs/alias_test.go
InTheCloudDan: I'm not sure, I'll leave it there for this PR just to keep it moving along.

coderefs/alias_test.go
InTheCloudDan: updated!

PR #148, opened 2021-05-25
doc: fix some bash script

chenrui333:


chenrui333:
I think this is a good fix for the repo. cc @InTheCloudDan @yusinto

vlymar (approved):
Thanks for the fix @chenrui333!

PR #153, opened 2021-06-14
change to use StringBuilder

InTheCloudDan:


yusinto (approved):

search/search.go
```diff
@@ -41,8 +41,13 @@ func MatchDelimiters(line, flagKey, delimiters string) bool {
 		return true
 	}
 	for _, left := range delimiters {
+		var sb strings.Builder
```
yusinto, line 44: Nit: perhaps we could add unit tests for this if there isn't any already?

InTheCloudDan: This should be covered by the current unit tests afaik, it's just changing the current implementation details.

PR #157, opened 2021-08-09
Refactor Code Refs to generic matcher

InTheCloudDan:
This PR creates a `Matcher` struct that contains Elements. A single element is a group of resources(currently flags) per project that code references is looking for. There are places in the code that are making the assumption that there is only a single element. I did it this way so there was not more refactoring because as of today all that is supported is one project per run.


element/element.go
```diff
@@ -0,0 +1,36 @@
+package element
```
ashanbrown, line 1: the name of this package isn't very illuminating.  Maybe just "matcher"?

element/element.go
```diff
@@ -0,0 +1,36 @@
...
+	Directory  string
+}
+
+type Matcher struct {
+	Elements   []ElementMatcher
+	CtxLines   int
+	Delimiters string
+}
+
+func (m Matcher) MatchElement(line, flagKey string) bool {
```
ashanbrown, line 19: This looks like it pre-supposes you want to do this work a flag key at a time.  I still question whether that is the right approach.  You could potentially take a list of keys.

ashanbrown, line 19: This looks like it pre-supposes you want to do this work a flag key at a time.  I still question whether that is the right approach.  You could potentially take a list of keys and return what keys it matches.  That might offer a more general solution.

element/element.go
```diff
@@ -0,0 +1,36 @@
+package element
+
+import "strings"
+
+type ElementMatcher struct {
```
ashanbrown, line 5: It might be useful to have constructor for this that can do any pre-processing work you might want to do.

element/element.go
InTheCloudDan: I split `element` package out from `matcher` because I was getting circular dependency issues with how everything is currently implemented and was trying to minimize how much refactoring was done.

element/element.go
InTheCloudDan: Similar to above I was trying to walk the line of how much to refactor and what could be done with subsequent PRs.

internal/helpers/helpers.go
```diff
@@ -11,3 +20,17 @@ func Dedupe(s []string) []string {
 	}
 	return ret
 }
+
+func MakeTimestamp() int64 {
+	return time.Now().UnixNano() / int64(time.Millisecond)
+}
+
+func FatalServiceError(err error, ignoreServiceErrors bool) {
+	if ld.IsTransient(err) {
```
yusinto, line 29: Nit: since this is checking for transient errors, maybe an alternative name is LogTransientError? Otherwise some may get confused the difference between `log.Error.Fatal` and `FatalServiceError`.

search/matcher.go
```diff
@@ -0,0 +1,80 @@
...
+	flagMatcher.Elements, flagMatcher.Aliases = flags.GenerateSearchElements(opts, repoParams)
+
+	matcher := Matcher{
+		Elements: []ElementMatcher{flagMatcher},
+		CtxLines: opts.ContextLines,
+	}
+
+	// Configure delimiters
+	delims := getDelimiters(opts)
+	matcher.Delimiters = strings.Join(helpers.Dedupe(delims), "")
```
yusinto, line 41: Does it make sense to detect duplicate delimiters early and fail early as part of config validation? I see we are already validating the configured directory in `NormalizeAndValidatePath`, we can expand this to include duplicate delimiters.

flags/flags.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	return filteredFlags, aliases
+}
+
+// Very short flag keys lead to many false positives when searching in code,
+// so we filter them out.
+func filterShortFlagKeys(flags []string) (filtered []string, omitted []string) {
+	filteredFlags := []string{}
+	omittedFlags := []string{}
```
yusinto, line 60: Nit: goland gave me warnings for these saying: `Empty slice declaration using a literal `. It made these suggestions which seem to make sense:

```go
   var filteredFlags []string
   var omittedFlags []string
```

coderefs/coderefs.go
```diff
@@ -16,20 +14,15 @@ import (
...
-func Scan(opts options.Options) {
+func Run(opts options.Options) {
 	dir := opts.Dir
 	absPath, err := validation.NormalizeAndValidatePath(dir)
 	if err != nil {
 		log.Error.Fatalf("could not validate directory option: %s", err)
 	}
 
 	log.Info.Printf("absolute directory path: %s", absPath)
+	ldApi := ld.InitApiClient(ld.ApiOptions{ApiKey: opts.AccessToken, BaseUri: opts.BaseUri, ProjKey: opts.ProjKey, UserAgent: "LDFindCodeRefs/" + version.Version})
```
yusinto, line 25: It seems `InitApiClient` returns a new `ApiClient` object every time. We call this init function again in `flags.go` [here](https://github.com/launchdarkly/ld-find-code-refs/blob/05b2b82f6253169a829d844c75a951148480dbec/flags/flags.go#L23) when doing the scan.  I don't think this is a big deal, but there's a chance we can re-use this first object and avoid having to re-initialise the client, which might give us a little perf boost.

flags/flags.go
InTheCloudDan: I just tried changing to what you suggested and also naming the return `filteredFlags` and `omittedFlags` both have caused tests to fail.

yusinto (approved):

search/matcher.go
InTheCloudDan: I'll note this down for a follow up PR. Given the size of this one currently if I can get it merged in these changes should be much easier.

coderefs/coderefs.go
InTheCloudDan: Same as above. I'll note this down for a follow up to investigate in a smaller PR. It's how it works today so for now leaving as is.

PR #162, opened 2021-09-23
Build map of delimiters once

InTheCloudDan:
sc119754

### Summary
* Update Code References so it only builds the map of delimiters a single time instead of repeatedly when scanning each line for flags.

### Testing
Updated test and added a new test for private function.

### DIFF
--- a/search/matcher.go
+++ b/search/matcher.go
@@ -85,7 +85,6 @@ func buildDelimiterList(flags []string, delimiters string) map[string][]string {
 		return delimiterMap
 	}
 	for _, flag := range flags {
-		//flagsDelimited := []string{}
 		tempFlags := []string{}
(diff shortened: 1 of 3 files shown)


search/matcher.go
```diff
@@ -64,17 +66,38 @@ func (m Matcher) MatchElement(line, flagKey string) bool {
...
 	return false
 }
+
+func buildDelimiterList(flags []string, delimiters string) map[string][]string {
+	delimiterMap := make(map[string][]string)
+	if delimiters == "" {
+		return delimiterMap
+	}
+	for _, flag := range flags {
+		//flagsDelimited := []string{}
```
jazanne, line 88:
```suggestion
```

jazanne (approved):
looks great

PR #166, opened 2021-10-27
docs: migrate formula to homebrew-core

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/88047

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,6 @@ All turn-key configuration methods (docker images used by services like CircleCI
 #### macOS
 
 ```bash
-brew tap launchdarkly/tap
 brew install ld-find-code-refs
 ```
 


carmenquan:
Hi @chenrui333 - thank you for contributing to our codebase! Currently our release pipeline relies on [GoReleaser](https://goreleaser.com/) which is why we rely on the `brew tap launchdarkly/tap` step in our docs. With your added formula into the homebrew-core project, this complicates the way we release coderefs and cannot currently support this change that you've requested.

Thanks again! If you have any suggestions, we're always listening and looking for ways to improve!
Carmen

chenrui333:
I think it is fine to keep the homebrew-core one, as it works for me and maybe others (I will help maintain that formula)

PR #171, opened 2021-11-04
Filter flags to a single environment

InTheCloudDan:
We do not need all of the environments returned when scanning for flags. This updates the query to first fetch a valid environment in the Project, then use that for subsequent requests.


jazanne (approved):
Nice change!

PR #176, opened 2021-11-12
Do not run extinctions with dry run

InTheCloudDan:
We do not currently have testing covering this. Will add a followup ticket for that.

### DIFF
--- a/coderefs/coderefs.go
+++ b/coderefs/coderefs.go
@@ -63,7 +63,7 @@ func Run(opts options.Options) {
 
 	handleOutput(opts, matcher, branch, repoParams, ldApi)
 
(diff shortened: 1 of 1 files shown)


jazanne (approved):
do you think we want the option to dry run the find extinctions function?

InTheCloudDan:
@jazanne interesting point. We might want to do that as it's own command similar to `Prune` then we could add a dry run to that? But extinctions is more about the keys.

I just pushed a commit with the logic changed, so it still outputs the number of flags found and lets you know. I'm open to rewording/changing

jazanne:
> do you think we want the option to dry run the find extinctions function?

this works for me!

PR #180, opened 2021-11-23
Add make command for updating github action

InTheCloudDan:
Add a single make command to increment the github action and update the container. This step is not part of the automated release yet. We can either add that into this PR or a separate followup.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,11 @@
+
 # Note: These commands pertain to the development of ld-find-code-refs.
 #       They are not intended for use by the end-users of this program.
 SHELL=/bin/bash
 GORELEASER_VERSION=v0.169.0
+LD_RELEASE_VERSION ?= 0.0.1-SNAPSHOT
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -71,7 +75,34 @@ RELEASE_CMD=curl -sL https://git.io/goreleaser | GOPATH=$(mktemp -d) VERSION=$(G
 publish:
 	$(RELEASE_CMD)
 
+GIT_COMMAND=git
```
yusinto, line 78: This looks good and it looks like it should work. However, it does bloat the Makefile, so I have two questions:
1. Should this live in a separate shell script?
2. If it makes it easier and more maintainable in the future, is it possible we implement this in node js?

What do you think?

yusinto: As discussed, let's move this to a separate script to keep the core Makefile small and go ahead with the merge. Thanks @InTheCloudDan !

yusinto (approved):

PR #181, opened 2021-12-01
Monorepo Support

InTheCloudDan:
Initial monorepo support. Mostly implemented in a backwards compatible way. Anyone running via CLI args today are supported.

Added:
* `project` block in `coderefs.yaml`. You can now supply a list of multiple projects, including their starting subdirectory and additional alias configurations specific to that project.
* Check so that `projKey` and new `projects` blocks cannot be used at the same time.

Changes:
* `accessToken` is used for all projects.
* `aliases` block is now global. If an alias is added there it will affect all projects.

Breaking Changes:
* When using `--outDir` we previously included the LaunchDarkly Project name in output file name, this has been removed because it may contain flags from multiple projects.

Testing:
Add a `coderefs.yaml` under `.launchdarkly` directory that contains multiple projects and optional dir along with aliases.
```
projects:
  - projectKey: code-refs-test
    dir: public/
    aliases:
       - type: camelcase
  - projectKey: demo-test-project

```

### DIFF
--- a/coderefs/coderefs.go
+++ b/coderefs/coderefs.go
@@ -3,6 +3,7 @@ package coderefs
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/launchdarkly/ld-find-code-refs/internal/git"
 	"github.com/launchdarkly/ld-find-code-refs/internal/helpers"
@@ -31,6 +32,7 @@ func Run(opts options.Options) {
 	branchName := opts.Branch
 	revision := opts.Revision
 	var gitClient *git.Client
+	var commitTime *time.Time
 	if revision == "" {
 		gitClient, err = git.NewClient(absPath, branchName, opts.AllowTags)
 		if err != nil {
@@ -38,6 +40,7 @@ func Run(opts options.Options) {
 		}
 		branchName = gitClient.GitBranch
 		revision = gitClient.GitSha
+		commitTime = gitClient.GitTimestamp
 	}
 
 	repoParams := ld.RepoParams{
@@ -63,6 +66,7 @@ func Run(opts options.Options) {
 		UpdateSequenceId: updateId,
 		SyncTime:         helpers.MakeTimestamp(),
 		References:       refs,
+		CommitTime:       commitTime,
 	}
 
 	handleOutput(opts, matcher, branch, repoParams, ldApi)
--- a/coderefs/coderefs.go
+++ b/coderefs/coderefs.go
@@ -113,16 +113,13 @@ func calculateStaleBranches(branches []ld.BranchRep, remoteBranches map[string]b
 
 func handleOutput(opts options.Options, matcher search.Matcher, branch ld.BranchRep, repoParams ld.RepoParams, ldApi ld.ApiClient) {
 	outDir := opts.OutDir
-	var projects []string
-	if len(opts.Projects) > 0 {
-		for _, proj := range opts.Projects {
-			projects = append(projects, proj.ProjectKey)
-		}
-	} else {
-		projects = append(projects, opts.AccessToken)
+	projectKeys := make([]string, 1)
+	for _, project := range opts.Projects {
+		projectKeys = append(projectKeys, project.ProjectKey)
 	}
+
 	if outDir != "" {
-		outPath, err := branch.WriteToCSV(outDir, repoParams.Name, opts.Revision, projects)
+		outPath, err := branch.WriteToCSV(outDir, repoParams.Name, opts.Revision)
 		if err != nil {
 			log.Error.Fatalf("error writing code references to csv: %s", err)
 		}
@@ -152,7 +149,7 @@ func handleOutput(opts options.Options, matcher search.Matcher, branch ld.Branch
 		branch.TotalHunkCount(),
 		len(matcher.Elements[0].Elements),
 		len(branch.References),
-		projects,
+		projectKeys,
 	)
 	err := ldApi.PutCodeReferenceBranch(branch, repoParams.Name)
 	switch {
--- a/coderefs/coderefs.go
+++ b/coderefs/coderefs.go
@@ -21,7 +21,7 @@ func Run(opts options.Options) {
 		})
 	}
 	dir := opts.Dir
-	absPath, err := validation.NormalizeAndValidatePath(dir)
+	absPath, err := validation.NormalizeAndValidatePath(opts.Dir)
 	if err != nil {
 		log.Error.Fatalf("could not validate directory option: %s", err)
 	}
@@ -50,7 +50,7 @@ func Run(opts options.Options) {
 		DefaultBranch:     opts.DefaultBranch,
 	}
 
-	matcher, refs := search.Scan(opts, repoParams)
+	matcher, refs := search.Scan(opts, repoParams, absPath)
 
 	var updateId *int
 	if opts.UpdateSequenceId >= 0 {
--- a/coderefs/coderefs.go
+++ b/coderefs/coderefs.go
@@ -20,7 +20,6 @@ func Run(opts options.Options) {
 			Key: opts.ProjKey,
 		})
 	}
-	dir := opts.Dir
 	absPath, err := validation.NormalizeAndValidatePath(opts.Dir)
 	if err != nil {
 		log.Error.Fatalf("could not validate directory option: %s", err)
--- a/coderefs/coderefs.go
+++ b/coderefs/coderefs.go
@@ -165,26 +165,31 @@ func handleOutput(opts options.Options, matcher search.Matcher, branch ld.Branch
 func runExtinctions(opts options.Options, matcher search.Matcher, branch ld.BranchRep, repoParams ld.RepoParams, gitClient *git.Client, ldApi ld.ApiClient) {
 	lookback := opts.Lookback
 	if lookback > 0 {
-		missingFlags := []string{}
-		for flag, count := range branch.CountByFlag(matcher.Elements[0].Elements) {
-			if count == 0 {
-				missingFlags = append(missingFlags, flag)
-			}
+		var removedFlags []ld.ExtinctionRep
+		for i, project := range opts.Projects {
+			missingFlags := []string{}
+			for flag, count := range branch.CountByFlag(matcher.Elements[i].Elements, project.ProjectKey) {
+				if count == 0 {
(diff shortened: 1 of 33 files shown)


coderefs/coderefs.go
```diff
@@ -15,14 +15,19 @@ import (
 )
 
 func Run(opts options.Options) {
+	if len(opts.ProjKey) > 0 {
```
InTheCloudDan, line 18: We set this inside run because all instances(github actions, bitbucket, CLI) call to `Run`. Converts legacy `projKey` into a `Project`.

coderefs/coderefs.go
```diff
@@ -15,14 +15,19 @@ import (
...
+	}
 	dir := opts.Dir
 	absPath, err := validation.NormalizeAndValidatePath(dir)
 	if err != nil {
 		log.Error.Fatalf("could not validate directory option: %s", err)
 	}
 
 	log.Info.Printf("absolute directory path: %s", absPath)
-	ldApi := ld.InitApiClient(ld.ApiOptions{ApiKey: opts.AccessToken, BaseUri: opts.BaseUri, ProjKey: opts.ProjKey, UserAgent: "LDFindCodeRefs/" + version.Version})
+	ldApi := ld.InitApiClient(ld.ApiOptions{ApiKey: opts.AccessToken, BaseUri: opts.BaseUri, UserAgent: "LDFindCodeRefs/" + version.Version})
```
InTheCloudDan, line 30: Only used to get the project for flags. Changed the calling function to pass in the relevant project.

coderefs/coderefs.go
```diff
@@ -108,8 +113,13 @@ func calculateStaleBranches(branches []ld.BranchRep, remoteBranches map[string]b
...
 func handleOutput(opts options.Options, matcher search.Matcher, branch ld.BranchRep, repoParams ld.RepoParams, ldApi ld.ApiClient) {
 	outDir := opts.OutDir
+	projectKeys := make([]string, 1)
+	for _, project := range opts.Projects {
+		projectKeys = append(projectKeys, project.ProjectKey)
+	}
+
 	if outDir != "" {
-		outPath, err := branch.WriteToCSV(outDir, opts.ProjKey, repoParams.Name, opts.Revision)
+		outPath, err := branch.WriteToCSV(outDir, repoParams.Name, opts.Revision)
```
InTheCloudDan, line 122: As noted in PR description. We no longer include the `projKey` in the output file name so this is a breaking change.

coderefs/coderefs.go
```diff
@@ -154,30 +168,31 @@ func runExtinctions(opts options.Options, matcher search.Matcher, branch ld.Bran
 	lookback := opts.Lookback
 	dryRun := opts.DryRun
 	if lookback > 0 {
-		missingFlags := []string{}
-		for flag, count := range branch.CountByFlag(matcher.Elements[0].Elements) {
-			if count == 0 {
-				missingFlags = append(missingFlags, flag)
-			}
+		var removedFlags []ld.ExtinctionRep
+		for i, project := range opts.Projects {
```
InTheCloudDan, line 172: Loop over all of the projects now to find extinctions. The API endpoint already accepts an array of extinctions and each extinction has its project key embedded in the struct.

flags/flags.go
```diff
@@ -15,35 +15,37 @@ const (
 	minFlagKeyLen = 3 // Minimum flag key length helps reduce the number of false positives
 )
 
-func GetFlagKeys(opts options.Options, repoParams ld.RepoParams) []string {
-	projKey := opts.ProjKey
+func GetFlagKeys(opts options.Options, repoParams ld.RepoParams) map[string][]string {
+	flagKeys := make(map[string][]string)
+	for _, proj := range opts.Projects {
```
InTheCloudDan, line 20: Performing the loop in here because it already is coupled pretty tightly to implementation. Down the line we could look to move this outside of the function.

internal/git/git.go
```diff
@@ -183,40 +185,49 @@ func (c Client) FindExtinctions(projKey string, flags []string, matcher search.M
 		if err != nil {
 			return nil, err
 		}
-		patch, err := changes.Patch()
+		patch, err := changes.PatchContext(context.Background())
```
InTheCloudDan, line 188: This allows us to access `FilePatches` below. So now we can loop over the files in the commit to see if they're in the subdirectory for that matcher.

internal/git/git.go
```diff
@@ -183,40 +185,49 @@ func (c Client) FindExtinctions(projKey string, flags []string, matcher search.M
...
-				if strings.HasPrefix(patchLine, "-") && !strings.HasPrefix(patchLine, "---") {
-					delta = 1
-				} else if strings.HasPrefix(patchLine, "+") && !strings.HasPrefix(patchLine, "+++") {
-					delta = -1
-				}
-				if delta != 0 && matcher.MatchElement(patchLine, flag) {
-					removalCount += delta
+		for _, filePatch := range patch.FilePatches() {
+			fromFile, toFile := filePatch.Files()
+			if project.Dir != "" && (toFile == nil || !strings.HasPrefix(toFile.Path(), project.Dir)) {
```
InTheCloudDan, line 194: Tested this logic with local changes. Need to add more test coverage, also would like to talk over the implementation.

internal/ld/ld.go
```diff
@@ -91,10 +91,10 @@ func InitApiClient(options ApiOptions) ApiClient {
 	}
 }
 
-func (c ApiClient) GetFlagKeyList() ([]string, error) {
+func (c ApiClient) GetFlagKeyList(projKey string) ([]string, error) {
```
InTheCloudDan, line 94: Mentioned above. Now we explicitly pass in the project instead of relying on the `options.ProjKey`

internal/ld/ld.go
```diff
@@ -551,23 +551,27 @@ func (t tableData) Swap(i, j int) {
...
-func (b BranchRep) CountByFlag(flags []string) map[string]int64 {
+func (b BranchRep) CountByFlag(flags []string, project string) map[string]int64 {
 	refCountByFlag := map[string]int64{}
 	for _, flag := range flags {
 		refCountByFlag[flag] = 0
 	}
 	for _, ref := range b.References {
 		for _, hunk := range ref.Hunks {
-			refCountByFlag[hunk.FlagKey]++
+			if hunk.ProjKey == project || project == "" {
```
InTheCloudDan, line 561: Allowing empty `""` for the testing below.

coderefs/coderefs.go
```diff
@@ -154,30 +168,31 @@ func runExtinctions(opts options.Options, matcher search.Matcher, branch ld.Bran
...
+			missingFlags := []string{}
+			for flag, count := range branch.CountByFlag(matcher.Elements[i].Elements, project.Key) {
```
InTheCloudDan, line 174: For how we generated and assign the matcher elements, there is one for each Project struct so we can use the index `i` here to look the corresponding values.

coderefs/coderefs.go
```diff
@@ -121,21 +131,25 @@ func handleOutput(opts options.Options, matcher search.Matcher, branch ld.Branch
 	}
 
 	if opts.DryRun {
+		totalFlags := 0
+		for _, searchElems := range matcher.Elements {
+			totalFlags += len(searchElems.Elements)
```
InTheCloudDan, line 136: Verify elements is correct number of flags.

flags/flags.go
```diff
@@ -15,35 +15,37 @@ const (
...
+		ldApi := ld.InitApiClient(ld.ApiOptions{ApiKey: opts.AccessToken, BaseUri: opts.BaseUri, UserAgent: "LDFindCodeRefs/" + version.Version})
+		isDryRun := opts.DryRun
 
-	ldApi := ld.InitApiClient(ld.ApiOptions{ApiKey: opts.AccessToken, BaseUri: opts.BaseUri, ProjKey: projKey, UserAgent: "LDFindCodeRefs/" + version.Version})
-	isDryRun := opts.DryRun
+		ignoreServiceErrors := opts.IgnoreServiceErrors
+		if !isDryRun {
+			err := ldApi.MaybeUpsertCodeReferenceRepository(repoParams)
```
InTheCloudDan, line 26: Can this be moved outside of project loop.

options/options.go
```diff
@@ -165,11 +171,20 @@ func (o Options) ValidateRequired() error {
...
-	if len(o.ProjKey) > maxProjKeyLength {
-		if strings.HasPrefix(o.ProjKey, "sdk-") {
-			return fmt.Errorf("provided projKey (%s) appears to be a LaunchDarkly SDK key", "sdk-xxxx")
-		} else if strings.HasPrefix(o.ProjKey, "api-") {
-			return fmt.Errorf("provided projKey (%s) appears to be a LaunchDarkly API access token", "api-xxxx")
+	if len(o.ProjKey) > 0 && len(o.Projects) > 0 {
+		return fmt.Errorf("`--projKey` cannot be combined with `projects` in configuration")
+	}
+
+	if len(o.ProjKey) > maxProjKeyLength && len(o.Projects) == 0 {
```
InTheCloudDan, line 178: remove `&& len(o.Projects) == 0` if `o.ProjKey` is less than maxProjKeyLength it shouldn't matter if projects is not empty.

yusinto:
This looks good! Apart from the dir ~ relative path issue we discussed in the [blitz](https://launchdarkly.atlassian.net/wiki/spaces/ENG/pages/1985446495/Monorepo+Blitz+-+Week+of+12+06) everything else seems to work!

yusinto (approved):

flags/flags.go
InTheCloudDan: fixed

InTheCloudDan:
@XieX I updated the sections of code we talked over. I also added a few new helper functions. Rather than looping over all of the references for each project in a loop multiple times we now return a `map[string]map[string]int64` which is `[project][flag]int` so it's less processing.

internal/git/git.go
InTheCloudDan: We added validation logic in a separate PR building on this one that the `project.Dir` has to be relative. Given that and how go-git returns paths as relative to the root of the repo this and the conditional inside it are checking if the path does not start the `projects` `dir`. If not then continue onto next patch

jazanne (approved):
i'm so excited about this change!

XieX (approved):

PR #188, opened 2021-12-08
Refactor event parsing

jazanne:
- return early to make function more readable
- update error message to only show tags
   ref in message when allowTags is true

### DIFF
--- a/build/package/github-actions/github-actions.go
+++ b/build/package/github-actions/github-actions.go
@@ -124,17 +124,26 @@ func parseBranch(ref string, event *Event, allowTags bool) (string, error) {
 	}
 	results := re.FindStringSubmatch(ref)
 
-	if results == nil {
-		// The GITHUB_REF wasn't valid, so check if it's a pull request and use the pull request ref instead
-		if event != nil && event.Pull != nil {
-			return event.Pull.Head.Ref, nil
-		}
-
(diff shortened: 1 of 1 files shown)


build/package/github-actions/github-actions.go
```diff
@@ -124,17 +124,26 @@ func parseBranch(ref string, event *Event, allowTags bool) (string, error) {
...
-		if allowTags && event != nil && event.Release != nil {
-			return event.Release.TagName, nil
-		}
+	// The GITHUB_REF wasn't valid, so check if it's a pull request and use the pull request ref instead
+	if event != nil && event.Pull != nil {
+		return event.Pull.Head.Ref, nil
+	}
+
+	// If it's not a pull request, check if it's a Release
+	if allowTags && event != nil && event.Release != nil {
```
InTheCloudDan, line 137: does `event.Release` also cover just creating tags without an associated release?

jazanne: no, but that should be caught in the block above with the `ref` which is used for push events - as far as i can tell a tag would only come into play from a `push` and `release` - anything you think i'm missing? i'll do another test though

InTheCloudDan: Yep that makes sense!

InTheCloudDan (approved):

PR #196, opened 2021-12-17
[sc-133063] Automate circleci orb publishing

yusinto:
Note that the prerequisite of this work is having a github owner circleci token. There is a [security review](https://launchdarkly.atlassian.net/browse/SEC-773) for this. In the meantime to test this, we can a dev version of the orb.

This pr migrates existing Circleci shell commands to built-in releaser steps:

1. (Done in past pr) Modify`update-version.sh` to update version numbers in circleci/orb.yml.
2. Modify `publish.sh` to install the circleci cli
3. Modify `publish.sh` to run circleci setup authenticating with a circleci token that has the github owner role.
4. Modify `publish.sh` to run circleci orb publish

### DIFF
--- a/.ldrelease/publish-circleci.sh
+++ b/.ldrelease/publish-circleci.sh
@@ -3,7 +3,8 @@
 # Run this in publish step after all version information have been updated.
 set -ev
 
-brew install circleci
+sudo apt install curl
+curl -fLSs https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh | sudo bash
 
 # Read 2 arguments from the command line so we can debug this script.
 # Argument 1 is the release version. Defaults to releaser env variable.
--- a/.ldrelease/publish-circleci.sh
+++ b/.ldrelease/publish-circleci.sh
@@ -4,7 +4,7 @@
 set -ev
 
 sudo apt install curl
-curl -fLSs https://raw.githubusercontent.com/CircleCI-Public/circleci-cli/master/install.sh | sudo bash
+./install-circleci
 
 # Read 2 arguments from the command line so we can debug this script.
 # Argument 1 is the release version. Defaults to releaser env variable.
--- a/.ldrelease/publish-circleci.sh
+++ b/.ldrelease/publish-circleci.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Run this in publish step after all version information have been updated.
+set -ev
+
+brew install circleci
+
+# Read 2 arguments from the command line so we can debug this script.
+# Argument 1 is the release version. Defaults to releaser env variable.
+RELEASE_VERSION=${1:-$LD_RELEASE_VERSION}
+
+# Argument 2 is the circleci token. Defaults releaser circleci secret.
+CIRCLECI_CLI_TOKEN=${2:-"${LD_RELEASE_SECRETS_DIR}/circleci_token"}
+CIRCLECI_CLI_HOST="https://circleci.com"
+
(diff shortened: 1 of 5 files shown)


.ldrelease/publish-circleci.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# Run this in publish step after all version information have been updated.
+set -ev
+
+brew install circleci
```
InTheCloudDan, line 6: I don't think this should be in the script. We can run a `which` to verify the user has it installed and error out if it isn't there

.ldrelease/publish-circleci.sh
```diff
@@ -0,0 +1,23 @@
...
+# Read 2 arguments from the command line so we can debug this script.
+# Argument 1 is the release version. Defaults to releaser env variable.
+RELEASE_VERSION=${1:-$LD_RELEASE_VERSION}
+
+# Argument 2 is the circleci token. Defaults releaser circleci secret.
+CIRCLECI_CLI_TOKEN=${2:-"${LD_RELEASE_SECRETS_DIR}/circleci_token"}
+CIRCLECI_CLI_HOST="https://circleci.com"
+
+# Validate circleci orb config. No publishing done on this step.
+circleci orb validate build/package/circleci/orb.yml || (echo "Unable to validate orb"; exit 1)
```
InTheCloudDan, line 17: We have a `make` version of this do we want to call that so if there's changes required we only need to do them in one spot.

.ldrelease/publish-circleci.sh
```diff
@@ -0,0 +1,23 @@
...
+
+# dev publish only, not production
+circleci orb publish build/package/circleci/orb.yml launchdarkly/ld-find-code-refs@dev:$RELEASE_VERSION --token $CIRCLECI_CLI_TOKEN --host $CIRCLECI_CLI_HOST
```
InTheCloudDan, line 20: Same as above, this should call the `make` command.

InTheCloudDan (changes requested):

.ldrelease/publish-circleci.sh
yusinto: We need the `circleci` cli installed in the releaser environment. If we don't install it, we won't be able to automate any of the circleci steps. I could be missing something obvious here, but I don't think installing the circleci cli is an issue?

InTheCloudDan: we should install it in the container if it's not there currently, `brew` is pretty mac dependent, while the releaser containers are using linux.

InTheCloudDan: looks like it's based on Ubuntu. So we can change it use `apt-get` and I think it should be good to go. Thinking on it more yeah this script is really only used in the release process.

InTheCloudDan: if we go this route though I think we'll need to install with curl to bash which isn't ideal https://circleci.com/docs/2.0/local-cli/#alternative-installation-method

yusinto: Good point re `brew` being osx specific. I've changed this now to use apt to install curl and then installing circleci using curl.

.ldrelease/publish-circleci.sh
yusinto: I think we should remove those circleci make targets and use this shell script to publish to circle. I've added a comment to do this when we have a github owner token but will leave them for now so we can continue publishing the old way. In the future, all publishing to circleci orbs should be done through `publish-circleci.sh`.

InTheCloudDan (approved):

PR #197, opened 2021-12-21
Add validation for subdirectories of projects

InTheCloudDan:
Adds basic validation of subdirectories via checking the directory is valid, and also making sure there are no preceding characters on the path such as `/`.

### DIFF
--- a/internal/validation/validation.go
+++ b/internal/validation/validation.go
@@ -45,3 +45,24 @@ func FileExists(filename string) bool {
 	}
 	return !info.IsDir()
 }
+
+func IsSubDirValid(base, subdir string) error {
+	if prefixExists(subdir) {
+		return fmt.Errorf(`project subdirectory should not start with prefix: %s`, string(subdir[0]))
+	}
+	path := filepath.Join(base, subdir)
+	pathInfo, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


internal/validation/validation.go
```diff
@@ -45,3 +45,24 @@ func FileExists(filename string) bool {
...
+
+	if !pathInfo.IsDir() {
+		return fmt.Errorf(`path: %s is not a directory`, path)
+	}
+
+	return nil
+}
+
+func prefixExists(path string) bool {
+	return path[0] == '\\' || path[0] == '/' || path[0] == '.'
```
jazanne, line 67: for `\\` would this need to be `return path[0] == '\'` OR `(path[0] == '\' && path[1] == '\')`

jazanne (approved):

internal/validation/validation.go
jazanne: but the test passes.... what am i missing? 😂

InTheCloudDan: I think I could change the delimiter but it wasn't happy with backticks iirc, but it's escaping the `\`

jazanne: AHHHHHHHHHHH thanks for the clarification! i probably would have realized that if i opened it in VSCode 🤦‍♀️

PR #198, opened 2021-12-22
Add timestamp of commit

InTheCloudDan:
We current support timestamp but it is unused. Additional updates will be needed server side, this is first step.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -21,9 +21,10 @@ import (
 )
 
 type Client struct {
-	workspace string
-	GitBranch string
-	GitSha    string
+	workspace    string
+	GitBranch    string
+	GitSha       string
+	GitTimestamp string
 }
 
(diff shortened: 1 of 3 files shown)


internal/git/git.go
```diff
@@ -21,9 +21,10 @@ import (
 )
 
 type Client struct {
-	workspace string
-	GitBranch string
-	GitSha    string
+	workspace    string
+	GitBranch    string
+	GitSha       string
+	GitTimestamp string
```
InTheCloudDan, line 27: Not sure if we should just leave this as `time.Time` instead of returning a string from the helper function.

jazanne: i'm all for keeping it time.Time instead of casting to string

InTheCloudDan: updated!

PR #206, opened 2022-01-05
match Extinction time rep

InTheCloudDan:
This is functionally the same as the current implementation for getting time but changed to use `int64` as it matches how Extinctions track their current time: https://github.com/launchdarkly/ld-find-code-refs/blob/master/internal/git/git.go#L243


jazanne (approved):
nice catch

PR #207, opened 2022-01-05
[sc-136606] Circleci publishing fixes

yusinto:
We've got our github owner token finally! Yay! This ticket is to fix a few issues found when publishing to circleci orbs using the new token:

* Fix ldreleaser path to the install cli script
* Uncomment production publish step
* Comment dev publish step
* Fix paths to other marketplace publish scripts

### DIFF
--- a/.ldrelease/publish-circleci.sh
+++ b/.ldrelease/publish-circleci.sh
@@ -4,7 +4,7 @@
 set -ev
 
 sudo apt install curl
-./install-circleci
+./.ldrelease/install-circleci
 
 # Read 2 arguments from the command line so we can debug this script.
 # Argument 1 is the release version. Defaults to releaser env variable.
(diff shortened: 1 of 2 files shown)


.ldrelease/publish-circleci.sh
```diff
@@ -4,7 +4,7 @@
 set -ev
 
 sudo apt install curl
```
yusinto, line 6: Locally this complains that java is not installed and broke the script because of the JDK install prompt. I hope on in the docker Ubuntu image this does not happen.

InTheCloudDan: Can we test this in the container locally before merging?

.ldrelease/publish-circleci.sh
```diff
@@ -3,8 +3,7 @@
 # Run this in publish step after all version information have been updated.
 set -ev
 
-sudo apt install curl
-./install-circleci
```
yusinto, line 7: @InTheCloudDan  I tested this in the [cimg/go](https://circleci.com/developer/images/image/cimg/go) container locally and it works. Turns out as a bonus that container already contains curl so we can remove the install curl command and execute the install directly.

InTheCloudDan (approved):

PR #211, opened 2022-01-18
Switch to use jitter with min/max

InTheCloudDan:
The current retries, try with a timed exponential backoff. If a branch has multiple commits that finish at roughly the same time this will cause them to all POST at the same time and if it errors to keep retrying in sync. This will cause them to have more variation between attempts. We still perform server side validation that the highest sync time is what is used.

### DIFF
--- a/internal/ld/ld.go
+++ b/internal/ld/ld.go
@@ -82,8 +82,8 @@ func InitApiClient(options ApiOptions) ApiClient {
 	if options.RetryMax != nil && *options.RetryMax >= 0 {
 		client.RetryMax = *options.RetryMax
 	}
-	client.RetryWaitMin = 1000 * time.Millisecond
-	client.RetryWaitMax = 30000 * time.Millisecond
+	client.RetryWaitMin = 1 * time.Second
+	client.RetryWaitMax = 30 * time.Second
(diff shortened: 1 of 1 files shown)


internal/ld/ld.go
```diff
@@ -80,6 +82,10 @@ func InitApiClient(options ApiOptions) ApiClient {
 	if options.RetryMax != nil && *options.RetryMax >= 0 {
 		client.RetryMax = *options.RetryMax
 	}
+	client.RetryWaitMin = 1000 * time.Millisecond
+	client.RetryWaitMax = 30000 * time.Millisecond
```
jazanne, line 86: TIL `time.Duration` is in nanoseconds o__O

...so my brain is malfunctioning 😅 

RetryWaitMin = 1 ms
RetryWaitMax = 30 ms

am i reading that right?

jazanne (approved):

internal/ld/ld.go
jazanne: my bad i was forgetting 0s. what if we did this to make it more clear?

```suggestion
	client.RetryWaitMin = 1 * time.Second
	client.RetryWaitMax = 30 * time.Second
```

PR #217, opened 2022-02-14
Fix only searching for a single extinction per project

InTheCloudDan:
* Add parameter to `Run` to determine if it should output the flag search(POST or generate CSV). This allows for new manual command to run extinctions only
* fix extinction bug related to count
* update test

### DIFF
--- a/coderefs/coderefs.go
+++ b/coderefs/coderefs.go
@@ -14,7 +14,7 @@ import (
 	"github.com/launchdarkly/ld-find-code-refs/search"
 )
 
-func Run(opts options.Options) {
+func Run(opts options.Options, output bool) {
 	if len(opts.ProjKey) > 0 {
 		opts.Projects = append(opts.Projects, options.Project{
 			Key: opts.ProjKey,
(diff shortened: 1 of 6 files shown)


coderefs/coderefs.go
```diff
@@ -68,7 +68,9 @@ func Run(opts options.Options) {
 		CommitTime:       commitTime,
 	}
 
-	handleOutput(opts, matcher, branch, repoParams, ldApi)
+	if output {
+		handleOutput(opts, matcher, branch, repoParams, ldApi)
+	}
```
jazanne, line 73: i understand that this is a function name that already existed, but it was reaaaaaally confusing to me for a few minutes how this enabled us to run just extinctions 😅  - can we rename it? Like, "saveReferences" or something?

jazanne (approved):
minor comment around function naming. This looks good

coderefs/coderefs.go
InTheCloudDan: @jazanne how about something like `generateHunkOutput` for naming? We should probably split this down more in the future and it's a relatively recent function, for right now though it does manage if it's writing to `csv` or `POST` to API so #namingthingsishard

jazanne: yea that works for me 👍

PR #224, opened 2022-02-28
Fixes slice bounds out of range error

jazanne:
* update truncateLine to count runes
* truncateLine maxCharCount is configurable
* add test cases

### DIFF
--- a/search/search.go
+++ b/search/search.go
@@ -5,6 +5,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"unicode/utf8"
 
 	"github.com/go-git/go-git/v5/plumbing"
 	"github.com/launchdarkly/ld-find-code-refs/internal/helpers"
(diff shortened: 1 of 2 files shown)


search/search.go
```diff
@@ -25,19 +26,13 @@ const (
 // Truncate lines to prevent sending over massive hunks, e.g. a minified file.
 // NOTE: We may end up truncating a valid flag key reference. We accept this risk
 //       and will handle hunks missing flag key references on the frontend.
-func truncateLine(line string) string {
-	// len(line) returns number of bytes, not num. characters, but it's a close enough
-	// approximation for our purposes
-	if len(line) <= maxLineCharCount {
+func truncateLine(line string, maxCharCount int) string {
```
jazanne, line 29: easier for testing, but also just a better design not to use a global constant within this function

search/search.go
```diff
@@ -25,19 +26,13 @@ const (
...
+	if utf8.RuneCountInString(line) <= maxCharCount {
```
jazanne, line 30: turns out this is faster than len([]runes) and it reads a lot better too

search/search_test.go
```diff
@@ -400,6 +400,46 @@ func Test_SearchForRefs(t *testing.T) {
...
+func Test_truncateLine(t *testing.T) {
+	tests := []struct {
+		name         string
+		line         string
+		maxCharCount int
+		want         string
+	}{
+		{
+			name: "line equal to max character count returns line",
+			// len(line) = 8
```
InTheCloudDan, line 412: Everything else looks good to me, can we just remove these commented lines if not needed?

InTheCloudDan (changes requested):

search/search_test.go
jazanne: i was keeping them in for reference about the diff between line length and rune length - what if i consolidate on one line?

PR #238, opened 2022-04-01
Add documentation for federal users

jazanne:
Adds notes about compiling with FIPS 140-2 support

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ This repository provides solutions for configuring [LaunchDarkly code references
 
 ## Turn-key Configuration options
 
-We provide turnkey support for common trigger mechanisms and CI / CD providers. You can also invoke the `ld-find-code-refs` utility from the command line, which can be run in any custom workflow you define (e.g. from a bash script, or a cron job).
+We provide turnkey support for common trigger mechanisms and CI/CD providers. You can also invoke the `ld-find-code-refs` utility from the command line, which can be run in any custom workflow you define, such as from a bash script or a cron job.
 
 | System           | Status                                                                                |
 | ---------------- | ------------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -99,6 +100,10 @@ docker run \
 
 Precompiled binaries for the latest release can be found [here](https://github.com/launchdarkly/ld-find-code-refs/releases/latest). Be sure to install the required [dependencies](#prerequisities) before running `ld-find-code-refs`.
 
+### Using code references in Federal environments
```
ember-stevens, line 103:
```suggestion
### Using code references in federal environments
```

README.md
```diff
@@ -28,15 +29,15 @@ This repository provides solutions for configuring [LaunchDarkly code references
 
 ## Turn-key Configuration options
 
-We provide turnkey support for common trigger mechanisms and CI / CD providers. You can also invoke the ld-find-code-refs utility from the command line, which can be run in any custom workflow you define (e.g. from a bash script, or a cron job).
+We provide turnkey support for common trigger mechanisms and CI / CD providers. You can also invoke the `ld-find-code-refs` utility from the command line, which can be run in any custom workflow you define (e.g. from a bash script, or a cron job).
```
ember-stevens, line 32:
```suggestion
We provide turnkey support for common trigger mechanisms and CI/CD providers. You can also invoke the `ld-find-code-refs` utility from the command line, which can be run in any custom workflow you define, such as from a bash script or a cron job.
```

ember-stevens (changes requested):

carmenquan (approved):

ember-stevens (approved):

README.md
```diff
@@ -99,6 +100,10 @@ docker run \
 
 Precompiled binaries for the latest release can be found [here](https://github.com/launchdarkly/ld-find-code-refs/releases/latest). Be sure to install the required [dependencies](#prerequisities) before running `ld-find-code-refs`.
 
+### Using code references in federal environments
+
+If you are using the FedRAMP compliant [LaunchDarkly federal instance](https://docs.launchdarkly.com/home/advanced/federal), the `ld-find-code-refs` binary should be compiled with FIPS 140-2 support by using a tool like [boringssl](https://github.com/golang/go/tree/dev.boringcrypto/misc/boring#caveat).
```
pkaeding, line 105: I don't think we need to link to the caveat anchor here; the full page should be relevant. Also `Go+BoringCrypto` seems to be what they call the project (boringssl is a slightly different project; this is the Golang integration of BoringSSL-- this is probably a distinction without a difference, but since this is public, we should try to call it by the official name).

```suggestion
If you are using the FedRAMP compliant [LaunchDarkly federal instance](https://docs.launchdarkly.com/home/advanced/federal), the `ld-find-code-refs` binary should be compiled with FIPS 140-2 support by using a tool like [Go+BoringCrypto](https://github.com/golang/go/tree/dev.boringcrypto/misc/boring).
```

pkaeding (approved):

README.md
jazanne: thanks for the feedback!

PR #262, opened 2022-08-31
[sc-45912] Add CI tool to request user agent

jazanne:
Example:

`LDFindCodeRefs/2.5.7 (CI github-actions)`

Caveat: it's an optional CLI parameter, so we can't require it when used with docker image (gitlab) or directly with tool

### DIFF
--- a/internal/ld/ld.go
+++ b/internal/ld/ld.go
@@ -18,7 +18,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/hashicorp/go-retryablehttp"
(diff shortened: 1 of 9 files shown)


tiffanylphan (approved):
👍

internal/ld/ld.go
```diff
@@ -18,7 +18,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/hashicorp/go-retryablehttp"
```
tiffanylphan, line 21: 🥇

PR #264, opened 2022-08-31
[sc-135747] use go git library for all git commands

jazanne:
Hoping that this will help in resolving issue w/new versions of git.

```
ERROR: 2022/08/31 18:27:52 coderefs.go:37: error parsing current commit sha: fatal: detected dubious ownership in repository at '/github/workspace'
To add an exception for this directory, call:

    git config --global --add safe.directory /github/workspace
```

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -122,15 +122,17 @@ func (c *Client) tagName() (string, error) {
 }
 
 func (c *Client) headSha() (string, error) {
-	/* #nosec */
-	cmd := exec.Command("git", "-C", c.workspace, "rev-parse", "HEAD")
-	out, err := cmd.CombinedOutput()
+	repo, err := git.PlainOpen(c.workspace)
 	if err != nil {
-		return "", errors.New(string(out))
+		return "", err
 	}
-	ret := strings.TrimSpace(string(out))
-	log.Debug.Printf("identified head sha: %s", ret)
-	return ret, nil
+	ref, err := repo.Head()
+	if err != nil {
+		return "", err
+	}
+	sha := ref.Hash().String()
+	log.Debug.Printf("identified head sha: %s", sha)
+	return sha, nil
 }
 
 func (c *Client) commitTime() (int64, error) {
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 
 	git "github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
 	object "github.com/go-git/go-git/v5/plumbing/object"
 
 	"github.com/launchdarkly/ld-find-code-refs/internal/ld"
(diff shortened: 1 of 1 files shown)


jazanne:
I can't figure out how to test that this fixes the specific issue locally, because i can't create the same issue. I tried nektos/act with [this configuration](https://github.com/launchdarkly/git-gatsby/pull/2094/files), but it doesn't set the permissions for files the same way, so it doesn't throw the error.

This might need to be a test in prod situation 🙃

internal/git/git.go
```diff
@@ -91,82 +91,112 @@ func (c *Client) getRef(branch string, allowTags bool) (name string, refType str
...
-func (c *Client) RemoteBranches() (map[string]bool, error) {
-	/* #nosec */
-	cmd := exec.Command("git", "-C", c.workspace, "ls-remote", "--quiet", "--heads")
-	out, err := cmd.CombinedOutput()
+func (c *Client) RemoteBranches() (branches map[string]bool, err error) {
+	branches = map[string]bool{}
+	repo, err := git.PlainOpen(c.workspace)
 	if err != nil {
-		return nil, errors.New(string(out))
+		return branches, err
```
InTheCloudDan, line 176: same as above question

InTheCloudDan (approved):
Only comments really, overall looks good!

internal/git/git.go
```diff
@@ -91,82 +91,112 @@ func (c *Client) getRef(branch string, allowTags bool) (name string, refType str
...
 }
 
-func (c *Client) branchName() (string, error) {
-	/* #nosec */
-	cmd := exec.Command("git", "-C", c.workspace, "rev-parse", "--abbrev-ref", "HEAD")
-	out, err := cmd.CombinedOutput()
+func (c *Client) branchName() (name string, err error) {
+	repo, err := git.PlainOpen(c.workspace)
+	if err != nil {
+		return name, err
```
mfreundlich, line 97: Isn't name undefined at this point? Or does go somehow know to magically make this the empty string. Same question for all the updated methods here that return tuples

jazanne: yep, it will just be an empty string

TIL, you don't even have to list the params in return statement :heart: https://go.dev/tour/basics/7

mfreundlich (approved):

internal/git/git.go
jazanne: (not going to update these function because they are a but longer/more complex than the example given)

internal/git/git.go
```diff
@@ -91,82 +91,112 @@ func (c *Client) getRef(branch string, allowTags bool) (name string, refType str
...
-		return nil, errors.New(string(out))
+		return branches, err
 	}
-	rgx := regexp.MustCompile("refs/heads/(.*)")
-	results := rgx.FindAllStringSubmatch(string(out), -1)
-	log.Debug.Printf("found %d branches on remote", len(results))
-	ret := map[string]bool{}
-	for _, r := range results {
-		ret[r[1]] = true
+	remote, err := repo.Remote("origin")
```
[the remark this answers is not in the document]
jazanne: good point... looking into it

internal/git/git.go
jazanne: same as above

internal/git/git.go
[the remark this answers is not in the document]
jazanne: @InTheCloudDan updated to handle all remotes - does that seem like a fair approach?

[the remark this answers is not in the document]
InTheCloudDan: I think so, I'm not totally sure on the logic for this section of the code it's not part of it that I've ever worked with before.

internal/git/git.go
```diff
@@ -91,82 +91,116 @@ func (c *Client) getRef(branch string, allowTags bool) (name string, refType str
...
+			return branches, err
+		}
+		refPrefix := "refs/heads/"
+		for _, ref := range refList {
+			refName := ref.Name().String()
+			if !strings.HasPrefix(refName, refPrefix) {
+				continue
+			}
+			branchName := refName[len(refPrefix):]
+			log.Debug.Printf("found remote branch: %s/%s", r.Config().Name, branchName)
```
jazanne, line 196: example...

```
DEBUG: 2022/09/01 10:27:38 git.go:196: found remote branch: origin/dob/testWindows
DEBUG: 2022/09/01 10:27:38 git.go:196: found remote branch: origin/dob/sc-129221/releaser
```

InTheCloudDan (approved):

PR #266, opened 2022-09-06
Add a missing backslash

suda:
The commit adding --userAgent argument broke the CircleCI orb due to a missing backslash


jazanne:
Thanks for this quick fix! We'll work on getting an updated package out shortly

PR #274, opened 2022-09-06
update go and docker image versions

jazanne:
- bumped go version in circleci
- bumped docker file versions
- had to update from precommit rules
- had to change install strategy for test deps in circleci workflow

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -13,6 +13,29 @@ linters:
       - lll
       - errcheck
       - interfacer
+      - tagliatelle
+      - varnamelen
+      - nonamedreturns
+      - nlreturn # re-enable this after fixing all cases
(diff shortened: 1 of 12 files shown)


ember-stevens (approved):

.golangci.yml
```diff
@@ -13,6 +13,29 @@ linters:
...
+      - wsl #fix
+      - gocognit
+      - revive
+      - errname
+      - cyclop
+      - gofumpt # fix
+      - gci
+      - nestif # fix
+      - gosec
+      - gocritic
```
jazanne, line 38: updated version of golangci-lint and had to disable some rules. we can revisit this later as KTLO to do lint fixes and re-enable these rules

tiffanylphan (approved):

Issue #311, opened 2023-01-06
orb not able to read parameters from bash_env


brondum:
We would like not to store our `LD_ACCESS_TOKEN` in our context or in the repo it self. Instead we pull this from our secret store and populate it to `$BASH_ENV` as a `pre_step`.   Currently the orb is running in `sh` instead of `bash` which means that CircleCI doesnt load bash_env automatically. 

Solutions: 
1. Could be to change shell to bash
2. source `$BASH_ENV` before the command

CircleCI docs explaining the use of env-vars and paramenters.
https://circleci.com/docs/env-vars/#parameters-and-bash-environment

> The $BASH_ENV workaround only works with bash, and has not been confirmed to work with other shells.

jazanne:
@brondum thank you for submitting this issue. We will have the team review the proposed change and get back to you.

jazanne:
@brondum thank you for your contribution fixing this issue. It is now available in [orb version 2.9.0](https://circleci.com/developer/orbs/orb/launchdarkly/ld-find-code-refs)

PR #312, opened 2023-01-06
adding source bash_env to load custom envs should solve #311

brondum:
This enable people to load secrets from envs instead of populating them in CONTEXTS or Projects Envs in CircleCI. Hence the latest CircleCI problems with access to their data.


brondum:
#311

jazanne:
@brondum thank you for submitting this PR. We will have the team review the change and get back to you.

build/package/circleci/orb.yml
```diff
@@ -115,6 +115,7 @@ jobs:
       - run:
           name: Find flag references
           command: |
+            source $BASH_ENV || true
```
pkaeding, line 118: Doesn't CircleCI [already do this](https://circleci.com/docs/env-vars/#parameters-and-bash-environment), automatically?

> In every step, CircleCI uses bash to source BASH_ENV. This means that BASH_ENV is automatically loaded and run, allowing you to use interpolation and share environment variables across run steps.

brondum: Only if the you use `bash` as shell and you specifically asks to use `sh`: 
https://github.com/launchdarkly/ld-find-code-refs/blob/main/build/package/circleci/orb.yml#L111

jazanne (approved):

PR #331, opened 2023-02-16
Fix: extinctions scans take hours

jazanne:
There was a logic issue that was causing a commit diff patch to be scanned multiple times - once for each changed file in the commit.

Bug introduced in [2.5.0](https://github.com/launchdarkly/ld-find-code-refs/releases/tag/2.5.0)


InTheCloudDan (approved):
Great win!

Issue #329, opened 2023-02-14
Error parsing git tag name when tag is annotated


kaarla:
I'm getting this error, but only when I use annotated tags:

```
INFO: 2023/02/10 16:17:10 coderefs.go:27: absolute directory path: /repo
DEBUG: 2023/02/10 16:17:10 git.go:105: identified branch name: HEAD
ERROR: 2023/02/10 16:17:10 coderefs.go:37: error parsing git tag name: git repo at /repo must be checked out to a valid branch or tag, or --branch option must be set
```
This is how the call looks like:
```#!/bin/sh -eo pipefail
ld-find-code-refs \
  --debug=true \
  --accessToken=${LD_ACCESS_TOKEN} \
  --projKey=default \
  --ignoreServiceErrors=false \
  --lookback=10 \
  --contextLines=2 \
  --baseUri=https://app.launchdarkly.com \
  --repoType=github \
  --repoUrl=https://github.com/particle-iot-inc/workerd \
  --defaultBranch=main \
  --allowTags=true \
  --commitUrlTemplate= \
  --hunkUrlTemplate= \
  --repoName=${CIRCLE_PROJECT_REPONAME} \
  --updateSequenceId=${CIRCLE_BUILD_NUM} \
  --dir=/repo \
  --userAgent="circle-ci"
```

I checked in the source code and never gets to this point:
https://github.com/launchdarkly/ld-find-code-refs/blob/4cbe7a942bfd2474c7522ae71311681054281c05/internal/git/git.go#L138

But it returns this message, which means it's not getting an error in the tagName() function:
https://github.com/launchdarkly/ld-find-code-refs/blob/4cbe7a942bfd2474c7522ae71311681054281c05/internal/git/git.go#L92

Checking the go-git library, don't find strong proofs to say that the function to retrieve or iterate over tags is only taking the lightweight tags, but the tests I made point to that. This is how my tags look like, 10 and 12 are the ones that work and the difference is those are lightweight while 9 and 11 are annotated:

[redacted hash] tag	refs/tags/8.3.9
[redacted hash] commit	refs/tags/8.3.10
[redacted hash] tag	refs/tags/8.3.11
[redacted hash] commit	refs/tags/8.3.12

jazanne:
@kaarla Sorry for the inconvience, we'll look into this issue.

[...]

PR #334, opened 2023-02-21
fix: failed to find tag name for annotated tags

jazanne:
For annotated tags, we need to check the [target `ref`](https://pkg.go.dev/github.com/go-git/go-git/v5@v5.5.2/plumbing/object#Tag) instead of ref of tag object

Fixes https://github.com/launchdarkly/ld-find-code-refs/issues/329

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -124,10 +124,11 @@ func (c *Client) tagName() (name string, err error) {
 	}
 
 	if err := iter.ForEach(func(ref *plumbing.Reference) error {
+		// Check if ref is an annotated tag
 		obj, err := repo.TagObject(ref.Hash())
 		if err != nil {
 			if errors.Is(err, plumbing.ErrObjectNotFound) {
-				// Lightweight tag
+				// Ref is lightweight tag
 				if head.Hash() == ref.Hash() {
 					name = ref.Name().Short()
 					iter.Close()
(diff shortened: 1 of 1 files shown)


internal/git/git.go
```diff
@@ -124,8 +124,21 @@ func (c *Client) tagName() (name string, err error) {
 	}
 
 	if err := iter.ForEach(func(ref *plumbing.Reference) error {
-		if head.Hash() == ref.Hash() {
-			name = ref.Name().Short()
+		obj, err := repo.TagObject(ref.Hash())
+		if err != nil {
+			if errors.Is(err, plumbing.ErrObjectNotFound) {
+				// Lightweight tag
```
annabkr, line 130: ooc: what is a lightweight tag?

internal/git/git.go
```diff
@@ -124,8 +124,21 @@ func (c *Client) tagName() (name string, err error) {
...
+				if head.Hash() == ref.Hash() {
```
annabkr, line 131: if this condition isn't met, do we want to return?

annabkr (approved):
left a couple questions but I trust your judgment

internal/git/git.go
jazanne: good point, thanks!

PR #338, opened 2023-03-01
Update CODEOWNERS with new PD org structure

alsmola:


jazanne (dismissed):
I think this should be @launchdarkly/team-accelerate-product instead

zmdavis:
@alsmola same deal here, looks like @launchdarkly/team-accelerate-product needs different permissions on this repo to be a valid codeowner (or maybe that error doesn't matter and we should ignore?)

alsmola:
> @alsmola same deal here, looks like @launchdarkly/team-accelerate-product needs different permissions on this repo to be a valid codeowner (or maybe that error doesn't matter and we should ignore?)

@zmdavis this should be resolved.

zmdavis (approved):

PR #343, opened 2023-04-04
add more info about executable alias files

jazanne:
Include commands for making a file executable when adding to git

### DIFF
--- a/docs/ALIASES.md
+++ b/docs/ALIASES.md
@@ -84,7 +84,7 @@ aliases:
   - type: filepattern
     paths:
       - '*[!_test].go'
-    patterns: 
+    patterns:
       - '(\w+) = "FLAG_KEY"'
 ```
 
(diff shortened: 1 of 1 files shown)


docs/ALIASES.md
```diff
@@ -97,13 +97,27 @@ Here's an example of a bash script which returns the the flag key as it's own al
 ```yaml
 aliases:
   - type: command
-    command: ./launchdarkly/launchdarklyAlias.sh # must be a valid shell command.
+    command: ./.launchdarkly/flagAlias.sh # must be a valid shell command.
     timeout: 5 # seconds
 ```
 
-Contents of `./launchdarkly/launchdarklyAlias.sh`:
+Contents of `./.launchdarkly/flagAlias.sh`:
```
jazanne, line 104: this is unrelated, but we have people add config.yaml to `.launchdarkly` directory, so i thought using that as an example makes more sense.

mmrj (approved):
Explanations LGTM. I did not try to run this.

PR #362, opened 2023-06-27
Update module-path to v2

jazanne:
module path should be updated for v2+

> module-path
The module's module path, usually the repository location from which the module can be downloaded by Go tools. For module versions v2 and later, this value must end with the major version number, such as /v2.

- [docs](https://go.dev/doc/modules/gomod-ref)

--

test with 

```
go get github.com/launchdarkly/ld-find-code-refs/v2@c5029a63b4da2a6ff9efbf63b400b16eff343d6c
```

### DIFF
--- a/search/search.go
+++ b/search/search.go
@@ -25,8 +25,7 @@ const (
 
 // Truncate lines to prevent sending over massive hunks, e.g. a minified file.
 // NOTE: We may end up truncating a valid flag key reference. We accept this risk
-//
-//	and will handle hunks missing flag key references on the frontend.
+// and will handle hunks missing flag key references on the frontend.
 func truncateLine(line string, maxCharCount int) string {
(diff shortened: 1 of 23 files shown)


search/search.go
```diff
@@ -25,7 +25,8 @@ const (
 
 // Truncate lines to prevent sending over massive hunks, e.g. a minified file.
 // NOTE: We may end up truncating a valid flag key reference. We accept this risk
-//       and will handle hunks missing flag key references on the frontend.
+//
+//	and will handle hunks missing flag key references on the frontend.
```
InTheCloudDan, line 29: @jazanne Seems like this is a typo? Otherwise looks good to me

InTheCloudDan (approved):
Other than what looks like a type lgtm

PR #370, opened 2023-07-27
[sc-210848] exclude flag key itself from alias list

jazanne:
For given flag key "test-flag" with configured `kebabcase` alias, own key "test-flag" should not be considered an alias

This will reduce scanning required.

### DIFF
--- a/aliases/alias.go
+++ b/aliases/alias.go
@@ -40,9 +40,23 @@ func GenerateAliases(flags []string, aliases []options.Alias, dir string) (map[s
 			}
 			ret[flag] = append(ret[flag], flagAliases...)
 		}
-		ret[flag] = helpers.Dedupe(ret[flag])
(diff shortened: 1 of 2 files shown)


aliases/alias.go
```diff
@@ -40,11 +40,25 @@ func GenerateAliases(flags []string, aliases []options.Alias, dir string) (map[s
 			}
 			ret[flag] = append(ret[flag], flagAliases...)
 		}
-		ret[flag] = helpers.Dedupe(ret[flag])
+		ret[flag] = uniqueAliases(flag, ret[flag])
 	}
 	return ret, nil
 }
 
+// Return a unique list of aliase, excluding flag key itself
```
jazanne, line 48:
```suggestion
// Return a unique list of aliases, excluding flag key itself
```

PR #372, opened 2023-08-03
update go to 1.20

jazanne:


annabkr (approved):
Does the `funlen` stuff need to be part of this? 

It seems like a pretty heavy handed lint rule to me 🤔 And it can't possibly be running in the `backend/` subdirectory because lord knows we have a million functions with more than 60 lines there.

jazanne:
> Does the funlen stuff need to be part of this?

no i'll disable funlen and move those commits elsewhere

PR #400, opened 2023-10-17
update requests made to use 'state' filter; request deprecated flags

cjhensen:
Update archived api call to use new state filter

### DIFF
--- a/internal/ld/ld.go
+++ b/internal/ld/ld.go
@@ -114,7 +114,7 @@ func (c ApiClient) GetFlagKeyList(projKey string) ([]string, error) {
 	}
 
 	flagReq := c.ldClient.FeatureFlagsApi.GetFeatureFlags(ctx, projKey).Summary(true)
-	archiveReq := c.ldClient.FeatureFlagsApi.GetFeatureFlags(ctx, projKey).Summary(true).Archived(true)
+	archiveReq := c.ldClient.FeatureFlagsApi.GetFeatureFlags(ctx, projKey).Summary(true).Filter("state:archived")
 
 	if len(project.Environments) > 0 {
 		// The first environment allows filtering when retrieving flags.


cjhensen:
Hey @jazanne! 👋🏻  Is there anything else that needs to be done here? Also - these requests depend on a feature flag. Is there a way of flagging these updates? Or would this not get merged until the flag is released in gonfalon to prod?

jazanne:
> Hey @jazanne! 👋🏻 Is there anything else that needs to be done here? Also - these requests depend on a feature flag. Is there a way of flagging these updates? Or would this not get merged until the flag is released in gonfalon to prod?

Hey, thanks for starting this PR! we'll also need to update the api version being called here https://github.com/launchdarkly/ld-find-code-refs/blob/530ed7683769b20ef0ad05648ea947d64cf04969/internal/ld/ld.go#L45

We don't have flags enabled in this tool, so I would say we should hold off on merging until the feature is released

internal/ld/ld.go
```diff
@@ -114,13 +114,15 @@ func (c ApiClient) GetFlagKeyList(projKey string) ([]string, error) {
 	}
 
 	flagReq := c.ldClient.FeatureFlagsApi.GetFeatureFlags(ctx, projKey).Summary(true)
-	archiveReq := c.ldClient.FeatureFlagsApi.GetFeatureFlags(ctx, projKey).Summary(true).Archived(true)
+	archiveReq := c.ldClient.FeatureFlagsApi.GetFeatureFlags(ctx, projKey).Summary(true).Filter("state:archived")
+	deprecatedReq := c.ldClient.FeatureFlagsApi.GetFeatureFlags(ctx, projKey).Summary(true).Filter("state:deprecated")
```
jazanne, line 118: @cjhensen sounds like we're going to defer this change, so i'd like to keep things to two requests for now, if you don't mind changing it - still would love to keep the change of using `Filter` instead of `Archived`

cjhensen: sure no problem, I'll update!

jazanne: thank you!

jazanne (approved):
looks good, we'll merge after release

jazanne (approved):

PR #408, opened 2023-11-03
perf: alias scanning in files

jazanne:
Improve performance of alias scanning in files in two ways

1. cache the results of `doublestar.FilepathGlob(glob)` so that consecutive calls with same pattern can use results
2. ~swap out golang regexp with https://github.com/wasilibs/go-re2~

With just change 1 I still found the tools running over 30 min (i stopped it before it finished) in one repo, but by adding change 2 it cut the time to under 10 min. Similarly, just implementing change 2 saw run times over 20 min (i stopped it before it finished).

This isn't a total fix for performance of scanning, depending on size of repo and wildcard patterns used, but it certainly helps.

UPDATE:
pulling back on the re2 usage since the build fails - we can revisit that later


aliases/testdata/wild/nested-wild/another/another/alias_test.txt
```diff
@@ -0,0 +1 @@
+EVEN_WILDER = 'wildFlag'
```
ld-kyee, line 1: 😂

ld-kyee (approved):
Can't say I understand entirely how this all works but the code seems 💯

mmrj (approved):
Approving to allow work to proceed. The Docs team is codeowners of `*.md` files in this repo, so are autotagged here, but I did not review the vendor `*.md` files carefully.

jazanne:
unfortunately the build is failing now that we're using re2, so i'm going to revert that change

Issue #351, opened 2023-05-15
Extinctions scan fails for large file diffs


ffluk3:
When running `ld-find-code-refs`, we see the following error:

```
panic: runtime error: index out of range [65533] with length 55989

goroutine 1 [running]:
github.com/sergi/go-diff/diffmatchpatch.(*DiffMatchPatch).DiffCharsToLines(0xc000fe4620?, {0xc00189e930, 0x1, 0x11000?}, {0xc00de3c000, 0xdab5, 0x11000?})
	/tmp/project-releaser/project/vendor/github.com/sergi/go-diff/diffmatchpatch/diff.go:452 +0x2bb
github.com/go-git/go-git/v5/utils/diff.DoWithTimeout({0xc00b036000?, 0x0?}, {0xc00ca1c000?, 0x800?}, 0x7f453cc8f108?)
	/tmp/project-releaser/project/vendor/github.com/go-git/go-git/v5/utils/diff/diff.go:37 +0x11d
github.com/go-git/go-git/v5/utils/diff.Do(...)
	/tmp/project-releaser/project/vendor/github.com/go-git/go-git/v5/utils/diff/diff.go:22
github.com/go-git/go-git/v5/plumbing/object.filePatchWithContext({0xbdad68, 0xc00013c000}, 0xc00706f200)
	/tmp/project-releaser/project/vendor/github.com/go-git/go-git/v5/plumbing/object/patch.go:68 +0x192
github.com/go-git/go-git/v5/plumbing/object.getPatchContext({0xbdad68, 0xc00013c000}, {0x0, 0x0}, {0xc00195adc0, 0x57, 0x0?})
	/tmp/project-releaser/project/vendor/github.com/go-git/go-git/v5/plumbing/object/patch.go:38 +0xf4
github.com/go-git/go-git/v5/plumbing/object.Changes.PatchContext(...)
	/tmp/project-releaser/project/vendor/github.com/go-git/go-git/v5/plumbing/object/change.go:158
github.com/launchdarkly/ld-find-code-refs/internal/git.Client.FindExtinctions({{0x7ffccf09be80, 0x5}, {0xc00013de88, 0x8}, {0xc0001258f0, 0x28}, 0x188209ab5b0}, {{0x7ffccf09bd4b, 0x7}, {0x0, ...}, ...}, ...)
	/tmp/project-releaser/project/internal/git/git.go:264 +0x54e
[...]

PR #386, opened 2023-08-24
fix: update go-diff package

ffluk3:
Resolves #351, where the tool would throw index out of range errors for sufficiently large source code diffs. The downstream library was recently updated to allow for significantly larger diff max sizes, as discussed in the associated issue.

PR #425, opened 2024-01-23
[sc-112730] fix: out of range error while running extinctions on large diffs

jazanne:
Thanks @ffluk3 for introducing fix in https://github.com/launchdarkly/ld-find-code-refs/pull/386

Adds a test case to verify the fix. The test data is taken from https://github.com/sergi/go-diff/issues/89 data.txt

Closes https://github.com/launchdarkly/ld-find-code-refs/issues/351


karayusuf (approved):

ffluk3:
Thank you @jazanne! Can we also publish a patch version for this change?

jazanne:
@ffluk3 yes, i expect to have that our later today

Issue #429, opened 2024-02-09
Multi-Project Scanning exits early when empty projects are found


arekmano:
### Problem

Multi-project scanning exits early when no valid flags are found in an individual project. When determining which feature flags to scan for, upon encountering a project without any flags, execution exits early ([link](https://github.com/launchdarkly/ld-find-code-refs/blob/d4d4e16cb740cfe26d43a99828c057610133ee2c/flags/flags.go#L40)).

This is somewhat unexpected, as the CLI tool already handles edge cases gracefully. For example, by filtering feature flag names that are less than 3 characters long.

### Impact

Our organization would like to rely on this tool to perform code scanning on projects. This bug is causing some issues that require hacky work-arounds with our current multi-project code scanning. Specifically, we now have to "pre-filter" the projects for those that contain feature flags, before using this tool. 

### Repro

1. Create a LaunchDarkly project with no feature flags
2. Create a LaunchDarkly project with a valid feature flag
3. Run the cli tool scanning for references using both projects.
4. Execution exits early 

### Solution

Rather than exiting early when an empty project is encountered, simply ignore that project, and log a warning

#### Additional Context

I am willing to contribute the code change, as long as I this is seen as a change that would be approved here.

jazanne:
@arekmano this is definitely an oversight, thanks for reporting! Feel free to submit a fix, we really appreciate it

arekmano:
A tentative PR was created to address this ^

jazanne:
Fixed in 2.11.8. thank you @arekmano

PR #430, opened 2024-02-09
fix(#429): Multi-project scans now ignore projects without valid flags

arekmano:
### Description

- Addresses issue #429.
- Projects without invalid flags will be ignored. A log message is written when this occurs.


jazanne (approved):
thanks for this fix!

PR #434, opened 2024-02-29
feat: respect X-Ratelimit-Reset when retrying API requests

### LINKED ISSUE #433: Handle API Rate Limit on call to `GetFlagKeyList`
## Current behavior

A call to `GetFlagKeyList` returns a `429 Too Many Requests` causes `ld-find-code-refs` to either exits with exit code 0 if `ignoreServiceErrors` is set to `true` or throws a fatal error.

## Expected behavior

A call to `GetFlagKeyList` returns a `429 Too Many Requests` is handled by retrying the request after the `X-Ratelimit-Reset` time has passed. A maximum number of retries is configured through a new argument. If the maximum number of retries is hit then an error is thrown.

I plan on working on this change since it's currently affecting about 10% of our CI runs.

jonathan-ostrander:
Fixes #433. Fallback to default behavior (the `RetryWaitMin` and `RetryWaitMax` lines were removed because the hardcoded values were already the default in `go-retryablehttp`) in the case where `X-Ratelimit-Reset` is not available for some reason.


jonathan-ostrander:
Just a friendly ping @jazanne since you approved [my coworker's](https://github.com/launchdarkly/ld-find-code-refs/commits?author=arekmano) PR a few weeks ago. Also, 👋 from Fishtown since I see you're also Philly-based.

jazanne:
hey @jonathan-ostrander thanks for this change! I'm working on getting it through the reviewing process. just out of curiosity, how are you all using code refs? via GitHub action, or the CLI directly?

jonathan-ostrander:
We use the GitHub Action. I was actually going to follow-up with a feature request issue and another PR for adding an option to scan all projects in an organization instead of a manually configured list. Right now we hack around this problem by having a step hit LD's API to list all of our projects and then we run the find code refs action for each of those. 

Currently it works alright, but ends up failing ~10% of the time because of rate limit issues.

jonathan-ostrander:
@jazanne Yes, we don't create LD projects _that_ often but when we do it would be nice to avoid the manual step of adding another project to a list in a GHA. That's a less important follow-up though. This change is a lot more helpful for us because currently ~10% of our actions are failing on the 429s.

internal/ld/ld.go
```diff
@@ -74,6 +74,32 @@ func IsTransient(err error) bool {
...
+							return time.Millisecond * time.Duration(sleep)
+						} else {
+							return time.Duration(0)
+						}
+					}
+				}
+			}
+		}
+
+		return h.DefaultBackoff(min, max, attemptNum, resp)
```
jazanne, line 99: Can you update this to previous behavior, which has `h.LinearJitterBackoff` as the default?

jazanne (changes requested):
Thanks for putting this together! After reviewing with the team, it's important that we keep the default retry strategy `LinearJitter` in place to handle cases where rate limit headers may not be available

internal/ld/ld.go
jonathan-ostrander: Done in 7d9f31f07a75805587000521a7b04963d42068e7. `h.DefaultBackoff` is still used for the unit test since `h.LinearJitterBackoff` depends on `math/rand`.

jazanne (approved):

PR #437, opened 2024-03-12
fix: use RateLimitBackoff client in LD API

jonathan-ostrander:
https://github.com/launchdarkly/ld-find-code-refs/pull/434 attempted to handle the rate limit from the LD API but the LD API client uses `http.DefaultClient` by default. This change passes in the client with `RateLimitBackoff` set into the call to init the LD API client.

### DIFF
--- a/internal/ld/ld.go
+++ b/internal/ld/ld.go
@@ -85,6 +85,7 @@ func RateLimitBackoff(now func() time.Time, fallbackBackoff h.Backoff) func(time
 				if s, ok := resp.Header["X-Ratelimit-Reset"]; ok {
 					if sleepUntil, err := strconv.ParseInt(s[0], 10, 64); err == nil {
 						sleep := sleepUntil - now().UnixMilli()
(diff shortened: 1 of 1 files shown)


jonathan-ostrander:
@jazanne sorry for needing a follow-up PR, but the original PR did not actually affect any of the LD API calls. This should have the intended behavior with regards to the API rate limits.

jazanne:
@jonathan-ostrander thanks for raising this issue. We've decided to remove the use of that secondary api client [here](https://github.com/launchdarkly/ld-find-code-refs/pull/438). Hope to have a release out by end of week. Thank you for your patience

jonathan-ostrander:
@jazanne could we consider this a hot fix for the current build while that PR gets through review? I'm worried that such a large PR might take longer than the next couple of days to be merged.

jonathan-ostrander:
Whoops, forgot to add the `Request` to the test fixture. Fixed in ff79d49.

jazanne (approved):

PR #438, opened 2024-03-12
remove use of ldclient

jazanne:
We had partially switched to using ld client for requests, but it's easier for us to maintain one client, so I'm removing it and using the main http client for consistency.

Other
- add a helper for all api paths
- update to latest ldapi client `v15`, and latest api version `20220603`
- disable `nilnil` in linter


options/options.go
```diff
@@ -260,7 +260,7 @@ func (o Options) Validate() error {
 	}
 
 	if o.Revision != "" && o.Branch == "" {
-		return fmt.Errorf(`"branch" option is required when "revision" option is set`)
+		return errors.New(`"branch" option is required when "revision" option is set`)
```
jazanne, line 263:
```suggestion
		return fmt.Errorf(`"branch" option is required when "revision" option is set`)
```

internal/ld/ld.go
```diff
@@ -42,10 +40,10 @@ type ApiOptions struct {
 }
 
 const (
-	apiVersion       = "20210729"
+	apiVersion       = "20220603"
```
InTheCloudDan, line 43: @jazanne this looks good to me, but I don't know the status of the API version now. Are we handling the pagination easily at this point?

InTheCloudDan (approved):

internal/ld/ld.go
jazanne: pagination still isn't required (forthcoming in "v4")

internal/ld/ld.go
```diff
@@ -21,14 +20,13 @@ import (
 	h "github.com/hashicorp/go-retryablehttp"
 	"github.com/olekukonko/tablewriter"
 
-	ldapi "github.com/launchdarkly/api-client-go/v7"
+	ldapi "github.com/launchdarkly/api-client-go/v14"
```
mmrj, line 23: do you want to switch to v15 while you're here?

mmrj (approved):
✅ from Docs to unblock you -- looks like we were tagged because there are `*.md` files in the `vendor` update, but I only looked at the non-`vendor` files. Let me know if there's anything you want specific Docs feedback on.

PR #441, opened 2024-03-26
feat: don't skip .github dir

jazanne:
allow scanning for flag keys used in github workflows with https://github.com/launchdarkly/gha-flags

User will need to have delimiters turned off or use proper aliasing to pick them up

### DIFF
--- a/search/files.go
+++ b/search/files.go
@@ -82,6 +82,10 @@ func readFiles(ctx context.Context, files chan<- file, workspace string) error {
 		// Skip directories, hidden files, and ignored files
 		if strings.HasPrefix(info.Name(), ".") || allIgnores.Match(path, isDir) {
 			if isDir {
+				// don't skip github dir
+				if strings.HasPrefix(info.Name(), ".github") {
+					return nil
+				}
 				return filepath.SkipDir
 			}
 			return nil
--- a/search/files_test.go
+++ b/search/files_test.go
@@ -18,7 +18,7 @@ func Test_readFiles(t *testing.T) {
 		switch file.path {
 		case "fileWithNoRefs":
 			assert.Equal(t, []string{"fileWithNoRefs"}, file.lines)
-		case "fileWithRefs":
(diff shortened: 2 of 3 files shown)


karayusuf (approved):

docs/ALIASES.md
```diff
@@ -123,3 +123,29 @@ For a file that has already been committed:
 ```sh
 git update-index --chmod=+x .launchdarkly/flagAlias.sh
 ```
+
+## Finding flags used in GitHub workflow files
+
+If you are using [launchdarkly/gha-flags](https://github.com/launchdarkly/gha-flags), your flag keys will not be wrapped in delimiters, so it is important to either have the proper aliases configured or disable delimiters.
```
mmrj, line 129:
```suggestion
If you are using [launchdarkly/gha-flags](https://github.com/launchdarkly/gha-flags), your flag keys will not be wrapped in delimiters, so it is important to either configure the proper aliases or disable delimiters.
```
super nit: active voice, parallel structure

docs/CONFIGURATION.md
```diff
@@ -136,6 +136,6 @@ delimiters:
 
 ## Ignoring files and directories
 
-All dotfiles and patterns in `.gitignore` and `.ignore` will be excluded by default.
+All dotfiles and patterns in `.gitignore` and `.ignore` will be excluded by default, except the `.github` directory. Flags may be referenced when using [launchdarky/gha-flags](https://github.com/launchdarkly/gha-flags). If you would like to skip scanning these file, add `.github` to one of the ignore files.
```
mmrj, line 139:
```suggestion
All dotfiles and patterns in `.gitignore` and `.ignore` will be excluded by default, except the `.github` directory. Flags may be referenced when using [launchdarky/gha-flags](https://github.com/launchdarkly/gha-flags). If you would like to skip scanning these files, add `.github` to one of the ignore files.
```

mmrj (approved):
minor docs suggestions, lgtm

PR #467, opened 2024-12-03
Add option to set custom path

asljiva1:
Added options for users to set custom path to .launchdarkly/coderefs.yaml file.
This can be very useful for monorepos and if many teams have many projects.


stasquatch:
Addressed in https://github.com/launchdarkly/ld-find-code-refs/pull/471 - closing out.

PR #495, opened 2025-04-17
feat: add option to filter archived feature flags

### LINKED ISSUE #496: Allow tool to skip fetching archived feature flags
Hi there, I'd like an option for the tool to skip fetching archived feature flags. Not only will this improve performance but I believe it would improve correctness in my use case.

I have a scenario where an archived feature flag has the same name as some variables in the code and the tool is incorrectly calling this a code reference. 

This is my `coderefs.yaml`:

```yaml
aliases:
  - type: command
    command: ./.launchdarkly/generate_aliases.sh
    timeout: 5
```

This is the `generate_aliases.sh`:

```sh
#!/bin/bash

read -r flagKey

FLAG_FILE="Shared/Utilities/FeatureFlagKey.swift"

line=$(grep -E "case[[:space:]]+[a-zA-Z0-9_]+[[:space:]]*=[[:space:]]*\"$flagKey\"" "$FLAG_FILE")

aliasName=$(echo "$line" | awk '{
  for (i = 1; i <= NF; i++) {
    if ($i == "case") {
      print $(i + 1)
      break
    }
  }
}' | tr -d ' ')

aliasName=${aliasName%%=*}

if [ -n "$aliasName" ]; then
  echo "[\"$flagKey\", \"FeatureFlagKey.$aliasName\", \"variation(forKey: .$aliasName)\"]"
else
  echo "[\"$flagKey\"]"
fi
```

Running the tool like this:
`ld-find-code-refs --debug --dryRun -t=$SOME_TOKEN -d="$(pwd)" -p=default -r=some-repo --ignoreServiceErrors --outDir=.`

I gave a try in https://github.com/launchdarkly/ld-find-code-refs/pull/495 to try and implement the change 🙂

khoaHyh:
Closes #496

### Changes

Added a CLI flag, `--skipArchivedFlags`, to skip fetching archived flags from the LaunchDarkly API using the [API docs](https://launchdarkly.com/docs/api/feature-flags/get-feature-flags) as reference.

#### Considerations

I considered adding the option to get `live` and `deprecated` flags, excluding `archived` flags. Seeing as the tool currently doesn't have any mention of filtering on `state`, I kept it straightforward.


blaqbern (approved):
Changes look good to me. We're having some trouble with our build at the moment, but I'll get this merged as soon as I can.

PR #497, opened 2025-04-22
dont publish junit tests on forks

blaqbern:
This is causing builds to fail because secrets are not passed through to workflows that run on forks (see Hosh's comment [here](https://launchdarkly.atlassian.net/browse/IDE-1459?focusedCommentId=323468))

We've got a bit of a backup going on with PRs in this repo. This change should help unblock them.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ jobs:
     steps:
       - name: Checkout Code
         uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29
-      
+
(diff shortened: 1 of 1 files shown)


kcacosta (approved):
does the white space formatting matter in this case?

hoshsadiq (approved):

blaqbern:
> does the white space formatting matter in this case?

no, I don't think it does because the extra whitespace existed before and didn't seem to cause any issues. I just noticed it and thought I would clean it up while I was in here.

PR #498, opened 2025-04-22
docs: add configuration docs for skipArchivedFlags

khoaHyh:
Forgot to add docs for `--skipArchivedFlags` in https://github.com/launchdarkly/ld-find-code-refs/pull/495 😅


blaqbern (approved):
I also missed this! Thanks for following up!

PR #504, opened 2025-05-19
update api version string

sloloris:


blaqbern (approved):

monsagri (commented):
Damn, that's a lot of changes for upgrading one of our packages 😂 

As long as we're not relying on any endpoints with breaking/pagination changes I guess this should be fine?

blaqbern:
> Damn, that's a lot of changes for upgrading one of our packages 😂
> 
> As long as we're not relying on any endpoints with breaking/pagination changes I guess this should be fine?

it's completely removing the LD api-client-go v15 and replacing it with v17. That's why the diff is so big. Not sure why we versioned the api client that way

PR #516, opened 2025-07-14
account for subdirectory in ref filepaths

blaqbern:
We were inadverntently ignoring the `subdirectory` option when creating `File` reps, which led to the tool search zero files when using the project-level `dir` property in `.launchdarkly/coderefs.yml`. The reason for this is that when using the `subdirectory` property, the `workspace` (this is the root directory that `ld-find-code-refs` searches under) becomes `<root>/subdirectory`, but the project-level `dir` property is relative to the repo root, so the tool would end up looking in `<root>/subdirectory/subdirectory` for files to scan which obviously doesn't exist. The fix here is to remove the `subdirectory` (if it exists) from  the `file`'s `path` property. Existing tests pass. New tests for the `subdirectory` use-case in progress, so I'll leave in draft for now.

### DIFF
--- a/search/files.go
+++ b/search/files.go
@@ -64,7 +64,7 @@ func readFileLines(path string) ([]string, error) {
 	return lines, nil
 }
 
-func readFiles(ctx context.Context, files chan<- file, workspace string) error {
+func readFiles(ctx context.Context, files chan<- file, workspace, subdirectory string) error {
 	defer close(files)
 	ignoreFiles := []string{".gitignore", ".ignore", ".ldignore"}
 	allIgnores := newIgnore(workspace, ignoreFiles)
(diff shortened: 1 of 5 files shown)


search/files.go
```diff
@@ -108,9 +108,20 @@ func readFiles(ctx context.Context, files chan<- file, workspace string) error {
...
 	return filepath.Walk(workspace, readFile)
 }
+
+func resolvePath(path, workspace, subdirectory string) string {
+	dir := workspace
+	if subdirectory != "" {
+		dir = strings.TrimSuffix(workspace, "/"+subdirectory)
+	}
+
+	return strings.TrimPrefix(path, dir+"/")
```
blaqbern, line 126: the `+"/"` part is necessary here to avoid `//` in the path. Because of that, we also need to remove the leading `/` from the subdirectory for the same reason. Kinda confusing 😞. I'll add a unit test for this function to make sure it always behaves how we expect.

search/files_test.go
```diff
@@ -33,25 +33,72 @@ func Test_readFiles(t *testing.T) {
 	})
 
 	t.Run("explicitly ignore .github files", func(t *testing.T) {
-		files := make(chan file, 8)
```
sloloris, line 36: did you mean to delete this test?

blaqbern: This was converted to the `"without subdirectoy option"` test case (starting on line 36 in green)

PR #521, opened 2025-07-23
[REL-4161] add logic to update the changelog

blaqbern:
Another follow up for the migration to GHA from Releaser for this repo.
Releaser took care of updating the changelog, and I missed that I would need to include that in the GHA as well. The update will take place only after the release of the GHA, CircleCI Orb and BitBucket Pipe have completed successfully.

---

### DIFF
--- a/scripts/release/commit-and-tag.sh
+++ b/scripts/release/commit-and-tag.sh
@@ -9,16 +9,34 @@ tag_exists() (
   git rev-parse "${RELEASE_TAG}" >/dev/null 2>&1
 )
 
+update_changelog() (
+  local ts=$(date +"%Y-%m-%d"))
+  local changelog_entry=$(cat << EOF
+## [$LD_RELEASE_VERSION] - $ts
+$CHANGELOG_ENTRY
+EOF
+  )
+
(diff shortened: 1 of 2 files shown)


scripts/release/commit-and-tag.sh
```diff
@@ -9,17 +9,35 @@ tag_exists() (
...
+## [$LD_RELEASE_VERSION] - $ts
+$CHANGELOG_ENTRY
+EOF
+  )
+
+  # insert the new changelog entry (followed by empty line) after line 4
+  # of CHANGELOG.md
+  sed -i "4r /dev/stdin" CHANGELOG.md <<< "$changelog_entry"$'\n'
+
+  cat CHANGELOG.md
```
blaqbern, line 24: This is for debugging. I'll remove after I have a sense of whether this works as expected. Can't really run this on push/pull_request since the changelog text will come from the Action input, and I want to test it in the same way it's going to be used. For context: GHA inputs are only available on `workflow_dispatch`, but `workflow_dispatch` is only available on the main branch.

sheenacarswell (approved):

PR #522, opened 2025-07-24
[REL-4161] fix syntax error; add a bit more logging

blaqbern:
Hopefully this will be the last one 😅.
Yesterday evening I added the CHANGELOG update which I had overlooked initially because Releaser did it behind-the-scenes. But there was an extra `)` in the code 🤦, so that's fixed here, and I'm also logging out the changelog so future releasers will be able to see the diff and the updated CHANGELOG when they run a dry run or actual release. There's also a minor refactor here. It's nothing big, but it makes the updates a bit harder to grok, so I'll try to comment in the code to explain what was changed.

---

### DIFF
--- a/scripts/release/commit-and-tag.sh
+++ b/scripts/release/commit-and-tag.sh
@@ -10,33 +10,36 @@ tag_exists() (
 )
 
 update_changelog() (
-  local ts=$(date +"%Y-%m-%d"))
+  local ts=$(date +"%Y-%m-%d")
   local changelog_entry=$(cat << EOF
 ## [$LD_RELEASE_VERSION] - $ts
 $CHANGELOG_ENTRY
 EOF
-  )
 
   # insert the new changelog entry (followed by empty line) after line 4
   # of CHANGELOG.md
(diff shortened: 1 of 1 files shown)


scripts/release/commit-and-tag.sh
```diff
@@ -45,5 +48,13 @@ else
...
   git push origin HEAD
   git push origin "${RELEASE_TAG}"
+)
+
+if [[ "$DRY_RUN" == "true" ]]; then
+  update_changelog
+  release_summary
+  echo "Dry run mode: skipping commit, tag, and push"
+else
+  commit_and_tag
```
blaqbern, line 58: The code from the else block was pulled into `commit_and_tag` to make this easier to read. The diff looks like a mess, but that's really all it is.

scripts/release/commit-and-tag.sh
```diff
@@ -10,33 +10,36 @@ tag_exists() (
 )
 
 update_changelog() (
-  local ts=$(date +"%Y-%m-%d"))
+  local ts=$(date +"%Y-%m-%d")
```
blaqbern, line 13: D'oh!

sloloris (approved):
fingers crossed!

PR #524, opened 2025-07-24
[REL-4161] Cblackburn/rel 4161/fix release action config

blaqbern:
Copying artifacts a temporary directory so they can be picked up by `ncipollo/release-action`. This is the last piece that was missing, I think. I ran a dry run successfully and the CHANGELOG, diff and artifacts all looked correct. Also a bit of minor cleanup here as this should be just about finished now.

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,21 +1,20 @@
 name: Release
 
 on:
-  pull_request
-  # workflow_dispatch:
-  #   inputs:
-  #     dryRun:
-  #       description: Perform a dry-run only
-  #       required: false
-  #       type: boolean
-  #     releaseVersion:
-  #       description: Next release version
-  #       required: true
-  #       type: string
-  #     changeLog:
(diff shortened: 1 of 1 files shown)


sheenacarswell (approved):
Approving as this seems plausible, but I'm not an expert here

.github/workflows/release.yml
```diff
@@ -60,13 +61,14 @@ jobs:
...
             ./scripts/release/publish.sh
           fi
-      - name: Commit changes and tag
-        run: |
-          ./scripts/release/commit-and-tag.sh
-      - name: Create Github release
+          ls -1a "$ARTIFACT_DIRECTORY"
+      - name: commit changes and tag
+        run: ./scripts/release/commit-and-tag.sh
+      - name: create Github release
```
blaqbern, line 67: The diff looks so weird here, but this is just lower-casing the names for consistency and some other minor cleanup/logging. The order is _not_ changing here, even though it kinda looks like it.

PR #547, opened 2025-08-04
chore: [REL-4161] thread homebrew-tap deploy key through to goreleaser

blaqbern:
I think we need to add the `LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY` as an env var in order to make it available to goreleaser. I think the reason this didn't work with the `ssh_key_by_repo` action is because the GHA itself is not accessing the homebrew-tap repo; it's goreleaser that's trying to make changes there. However, since we added `homebrew-tap` as a `cross_repo_deploy_key` for `ld-find-code-refs`, I believe the secret lives on the repo (see the description in screenshot), so I'm hopeful grabbing it from `secrets` and setting as an env var will be enough to give access to goreleaser.

Screenshot 2025-08-04 at 1 29 55 PM

---

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,7 @@ jobs:
       DRY_RUN: ${{ inputs.dryRun || 'false' }}
       CHANGELOG_ENTRY: ${{ inputs.changeLog }}
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      HOMEBREW_GH_TOKEN: ${{ toJSON(secrets.LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY) }}
       ARTIFACT_DIRECTORY: "/tmp/release-artifacts"
     steps:
       - uses: actions/checkout@v4
@@ -43,7 +44,6 @@ jobs:
           repo_keys_map: |
             {
               "launchdarkly/find-code-references": ${{ toJSON(secrets.LAUNCHDARKLY_FIND_CODE_REFERENCES_DEPLOY_KEY) }},
-              "launchdarkly/homebrew-tap": ${{ toJSON(secrets.LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY) }}
             }
       - name: build
         run: |
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,6 +1,5 @@
 builds:
-  -
-    binary: ld-find-code-refs
+  - binary: ld-find-code-refs
     id: ld-find-code-refs
     main: ./cmd/ld-find-code-refs/
     env:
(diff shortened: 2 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -43,7 +44,6 @@ jobs:
           repo_keys_map: |
             {
               "launchdarkly/find-code-references": ${{ toJSON(secrets.LAUNCHDARKLY_FIND_CODE_REFERENCES_DEPLOY_KEY) }},
-              "launchdarkly/homebrew-tap": ${{ toJSON(secrets.LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY) }}
```
blaqbern, line 46: This action doesn't need access to `homebrew-tap`, so we don't actually need this.

.goreleaser.yml
```diff
@@ -39,44 +35,38 @@ builds:
...
       name: homebrew-tap
-      token: "{{ .Env.LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY }}"
-      # git:
-      #   url: git@github.com:launchdarkly/homebrew-tap.git
-      #   private_key: "{{ .Env.LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY }}"
-    folder: Formula
+      branch: main
+      git:
+        url: git@github.com:launchdarkly/homebrew-tap.git
+        private_key: "{{ .Env.HOMEBREW_GH_TOKEN }}"
```
blaqbern, line 68: This is the main change to this file. All other changes are just getting rid of some unorthodox formatting.

Makefile
```diff
@@ -72,7 +72,7 @@ clean:
 	rm -f build/package/github-actions/ld-find-code-refs-github-action
 	rm -f build/package/bitbucket-pipelines/ld-find-code-refs-bitbucket-pipeline
 
-RELEASE_CMD=curl -sL https://git.io/goreleaser | GOPATH=$(mktemp -d) VERSION=$(GORELEASER_VERSION) GITHUB_TOKEN=$(GITHUB_TOKEN) LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY=${LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY} bash -s -- --clean --debug --release-notes $(RELEASE_NOTES)
+RELEASE_CMD=curl -sL https://git.io/goreleaser | GOPATH=$(mktemp -d) VERSION=$(GORELEASER_VERSION) GITHUB_TOKEN=$(GITHUB_TOKEN) bash -s -- --clean --debug --release-notes $(RELEASE_NOTES)
```
blaqbern, line 75: removing the key here, since we're adding it to the repo itself in .goreleaser.yml

ldhenry (approved):

.goreleaser.yml
blaqbern: I think the issue was that the `LAUNCHDARKLY_HOMEBREW_TAP_DEPLOY_KEY` is not an env var. It lives only in the container that's running the action. So I'm adding it as a env var to see if that will get us there. 🤞

PR #554, opened 2025-08-06
chore: [REL-4161] use printf to ensure we get the exact value for homebrew key

blaqbern:
The saga continues. Last time I got an [error](https://github.com/launchdarkly/ld-find-code-refs/actions/runs/16777430844/job/47506788949#step:7:633) that is likely related to a malformed homebrew-tap deploy key. Giving `printf`, and no json encoding a shot. Hopefully, this will write the unaltered value of the key to the file and we'll be good to go.

---

### DIFF
--- a/scripts/release/stage-artifacts.sh
+++ b/scripts/release/stage-artifacts.sh
@@ -10,7 +10,7 @@ stage_artifacts() (
   # write homebrew key to temporary file for Goreleaser
   if [[ -n "${HOMEBREW_GH_TOKEN:-}" ]]; then
     HOMEBREW_KEY_PATH="/tmp/homebrew-tap-deploy-key"
-    echo "$HOMEBREW_GH_TOKEN" > "$HOMEBREW_KEY_PATH"
+    printf '%s\n' "$HOMEBREW_GH_TOKEN" > "$HOMEBREW_KEY_PATH"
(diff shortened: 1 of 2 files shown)


scripts/release/stage-artifacts.sh
```diff
@@ -10,7 +10,7 @@ stage_artifacts() (
   # write homebrew key to temporary file for Goreleaser
   if [[ -n "${HOMEBREW_GH_TOKEN:-}" ]]; then
     HOMEBREW_KEY_PATH="/tmp/homebrew-tap-deploy-key"
-    echo "$HOMEBREW_GH_TOKEN" > "$HOMEBREW_KEY_PATH"
+    printf '%s\n' "$HOMEBREW_GH_TOKEN" > "$HOMEBREW_KEY_PATH"
     chmod 600 "$HOMEBREW_KEY_PATH"
```
sheldonmartin, line 14: > Last time I got an [error](https://github.com/launchdarkly/ld-find-code-refs/actions/runs/16777430844/job/47506788949#step:7:633) that is likely related to a malformed homebrew-tap deploy key.

Is this the correct error? The error I'm seeing in the link above seems to point to a permissions issue. Permissions are set too 600 which should be fine assuming what ever created the file is also reading the file. If not, maybe we need to tweak that?

sheldonmartin (approved):

PR #558, opened 2025-08-07
chore: fix up tag checks

blaqbern:
Two things being updated here:
1. update the tag-check (for this repo) to only check the remote repo for the tag. It will definitely exist in the local copy since we need to create it before running Goreleaser. Other minor tweak is that instead of bailing on the entire action, we just return 0, so the action can continue to run and complete the final step of creating the Github release. The only reason we're likely to be running the action again if the tag was created, is that the Github release failed, so we want to let the action continue, so it can do that last step.
2. checking for existing tag (GHA and Bitbucket Pipe) on dry-run as well. This was fixed for real runs of the action, but dry-runs will also bomb out if the tag exists (the tag could exist if we previously had a partially successful run, but need to re-run the action to retry the parts that didn't complete successfully)

### DIFF
--- a/scripts/release/push-to-origin.sh
+++ b/scripts/release/push-to-origin.sh
@@ -2,21 +2,20 @@
 
 set -euo pipefail
 
-release_tag="v${LD_RELEASE_VERSION}"
+release_tag="v$LD_RELEASE_VERSION"
 
 tag_exists() (
-  git fetch --tags
-  git rev-parse "${release_tag}" >/dev/null 2>&1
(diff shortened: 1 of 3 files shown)


scripts/release/push-to-origin.sh
```diff
@@ -2,21 +2,20 @@
 
 set -euo pipefail
 
-release_tag="v${LD_RELEASE_VERSION}"
+release_tag="v$LD_RELEASE_VERSION"
```
blaqbern, line 5: just removing unnecessary curlies

cjhensen (approved):

PR #599, opened 2026-06-03
chore: remove SECURITY.md (repo is no longer public)

kparkinson-ld:
## Summary

Removes `SECURITY.md` entirely since this repo no longer has a public mirror and is only visible to LD employees. The file's security reporting instructions are unnecessary for an internal-only repository.

Link to Devin session: https://app.devin.ai/sessions/8a2f1b59831d41a3b2d5f9cdb844c79d
Requested by: @kparkinson-ld

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,5 +1,8 @@
 # Reporting and Fixing Security Issues
 
-Please report all security issues to the LaunchDarkly security team by submitting a bug bounty report to our [HackerOne program](https://hackerone.com/launchdarkly?type=team). LaunchDarkly will triage and address all valid security issues following the response targets defined in our program policy. Valid security issues may be eligible for a bounty.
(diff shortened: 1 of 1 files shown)


ember-stevens:
I wonder if we need this file at all, since we no longer have a public mirror of our repo? Everything in this repo is now visible only to LD employees.

kparkinson-ld:
> I wonder if we need this file at all, since we no longer have a public mirror of our repo? Everything in this repo is now visible only to LD employees.

If there's no public mirror and this repo will remain internal then probably not. I'll remove it instead.

kparkinson-ld:
@ember-stevens removed the `SECURITY.md` file from this repo, since it is no longer needed. Ready for your re-review.

ember-stevens (approved):
Thank you!

PR #491, opened 2025-04-12
perf: (issue #377) add caching and use re2 for alias scanning in files

khoaHyh:
Attempting to continue the work done in https://github.com/launchdarkly/ld-find-code-refs/pull/408. Closes #377.

### Changes

- use caching to optimize file globbing and regex compilation
- swap stdlib `regexp` out for [go-re2](https://github.com/wasilibs/go-re2) for improved regex compilation and pattern matching
- added a test to assert alias-finding logic works properly when given a set of files with flag references in them

### Benchmarks

I compiled and executed locally using `local-find-code-refs --debug --dryRun -t=$SOME_TOKEN -d=. -p=default -r=some-repo --ignoreServiceErrors`

Ran this on a repo where the command execution never completed (waited >30 minutes until I cancelled) and when using this branch's version, the command was able to complete within 8m14s.

Issue #377, opened 2023-08-10
Tool gets stuck when defining glob filepattern on big repo


kostyay:
Hey
This is my config:
```yaml
aliases:
  - type: kebabcase
  - type: filepattern
    paths:
      - '**/manifest.yaml'
    patterns:
      - 'featureFlag: FLAG_KEY'
      - 'feature_flag: FLAG_KEY'
```

Running the tool like this:
```
ld-find-code-refs -t KEY -d ./ -p default -r REPO --debug -b master -T github -B master --dryRun -C 0 -l 1
```

Get the following output and the tool gets stuck:
```
INFO: 2023/08/10 20:15:58 coderefs.go:27: absolute directory path: XXXX
INFO: 2023/08/10 20:15:58 git.go:48: git branch: master
DEBUG: 2023/08/10 20:15:58 git.go:169: identified head sha: XXXXX
```

Could it be that I got rate limited because I was running it a bunch of times?
There are 317 files matching that pattern in the repo.

Sometimes it works then I get this output:
```dry run found 1113 code references across 720 flags and 844 files```

However its still not finding everything I'd expect to be found.
Is there a more verbose debugging mode?

jazanne:
@kostyay unfortunately, it is a known issue that the doublestar pattern does not perform very well for alias searching. We recommend using a more limited file pattern if possible. Given that you have 317 patterns to match in the repo, i understand that could be difficult.

There isn't more verbose debugging at this time for alias searching.

kostyay:
Ive tried using 
`xxx/*/*/manifest.yaml` and it didnt help either

kostyay:
@jazanne 
You can significantly speed up the initial parsing by caching the results returned by this function:
https://github.com/launchdarkly/ld-find-code-refs/blob/68b13f09c742e421f552fbe3c085a6df2f68bdd1/aliases/alias.go#L79

Example:
```go
var filepathGlobCache = map[string][]string{}

func filepathGlob(absGlob string) ([]string, error) {
	cachedFilePaths, ok := filepathGlobCache[absGlob]
	if ok {
		return cachedFilePaths, nil
	}

	matches, err := doublestar.FilepathGlob(absGlob)
	filepathGlobCache[absGlob] = matches

	return matches, err
}```

khoaHyh:
[...]

PR #604, opened 2026-07-07
fix: concatenate filepattern alias contents once per alias instead of once per flag

jazanne:
Disclaimer: I used AI to generate this PR and fix - if that's not allowed please let me know.

## Problem

`GenerateAliases` calls `GenerateAliasesFromFilePattern` once per flag key, and that function rebuilds the full concatenation of all glob-matched file contents (plus a `[]byte` → `string` copy for the regex engine) on every call.

On a repo with ~80MB of matched source and ~160 flags, peak allocation reaches tens of GB and the process is OOM-killed (exit 137) before producing any output. The same blowup affects find-code-references-in-pull-request, which consumes this package as a library and passes all flag keys through the same path.

## Fix

Hoist the concatenation out of the per-flag loop: `GenerateAliases` now builds each filepattern alias's contents once and reuses it across flags. `GenerateAliasesFromFilePattern` keeps its public signature and behavior, delegating to the extracted `concatFilePatternContents` + `matchFilePatternAliases` helpers.

With this change the same scan (~160 flags, ~80MB matched contents, alpine container) completes in ~14s with no OOM.

## Testing

```
go test ./aliases/...
```

Existing suite passes unchanged; no behavior change, output is identical for the repos we compared before/after.

## Prior art

#491 (for #377, "Tool gets stuck when defining glob filepattern on big repo") added glob and regex caching on this same path, which addressed the CPU side of big-repo filepattern scans. This change addresses the remaining memory side: the per-flag rebuild of the concatenated file contents.


cspath1 (approved):
Thank you for this! It looks like there's a minor lint issue with `allFileContents` being unused now. Approving now so it can be merged once that's been fixed!

cspath1 (approved):

